Issue #2, opened 2026-07-28
Setup-screen "Copy setup link" button gets stuck on "Copied!" after rapid double-click


jsundram:
## Summary

The setup/login screen's **Copy setup link** button (`handleCopyMobileLink` in `src/app.js`) has a "Copied!" flash that can get permanently stuck if the button is clicked twice within the ~1.5s flash window.

## Cause

The flash captures the button's *current* text as the "original" to restore:

```js
const original = btn.text().trim();
btn.text('Copied!');
setTimeout(() => btn.text(original), 1500);
```

On a rapid second click, `btn.text()` already reads `"Copied!"`, so that becomes the saved "original". When the timers fire, the label is restored to `"Copied!"` and never returns to `"Copy setup link"`.

## Fix

The in-app hamburger-menu version of this button was already made re-entrancy-safe via the `_flashLabel()` helper (preserves the true original text and resets the pending timer on each call — see commit `a21ee2d`). The setup-screen button should be routed through the same helper (or given equivalent guarding) so the two behave consistently.

Note `_flashLabel` currently targets a d3 selection's text; the setup button also toggles `#setupError` visibility, so a small amount of adaptation is needed rather than a drop-in swap.

## Repro

1. Open the setup screen with a valid CSV URL entered.
2. Double-click **Copy setup link** quickly.
3. Observe the button stays on "Copied!" instead of reverting.

## Severity

Cosmetic — copying still works; only the label label is wrong until the view is rebuilt.

PR #4, opened 2026-08-07
add max-streak metric to aggregate stats, calendar, and dashboard

jsundram:
Introduce a fifth metric — longest run of consecutive days with at least
one whole piece logged — alongside pieces / unique pieces / unique people /
playing days.

- computeAggregateStats now returns maxStreak, computed via a new
  longestConsecutiveRun helper over DST-safe day ordinals (Date.UTC on the
  local Y/M/D). The streak is scoped to the passed-in slice, so it obeys the
  dashboard filters and the "last 365 days" window.
- Calendar "Last 365 days" block adopts the dashboard's short labels
  (Pieces / Unique / People / Days) so all five fit on one line, and adds
  Streak. Per-year stats column adds a fifth "Longest streak in <year>"
  number in both the horizontal and vertical/fullscreen layouts, split at
  the year boundary.
- Dashboard KPI tiles and the Home ALL tab both surface the new metric.
- Tests cover runs, gaps, same-day dedup, isolated days, a spring-forward
  DST boundary, and the longestConsecutiveRun helper directly.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01A3x1bftDa2ZPigS3yCoLSW

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,7 +67,8 @@ Hash routing lives in `NavigationComponent`: menu clicks set `window.location.ha
   - `parseWork`, `processRow`, `fillForward`, `createEmptyRow`
   - `normalizePlayerNames` (applies `PLAYER_ALIASES` per slot class)
   - `peopleKeysFor(d)` — canonical-name keys for unique-people counting
-  - `computeAggregateStats(rows)` — `{ pieces, uniquePieces, uniquePeople, daysPlayed }`; used by both Calendar's "Last 365 days" and the ALL tab
+  - `computeAggregateStats(rows)` — `{ pieces, uniquePieces, uniquePeople, daysPlayed, maxStreak }`; used by Calendar's "Last 365 days", the Dashboard KPI tiles, and the ALL tab. `maxStreak` is the longest run of consecutive playing days (via `longestConsecutiveRun` over DST-safe day ordinals), scoped to the passed-in slice
+  - `longestConsecutiveRun(days)` — longest run of consecutive integer day ordinals in a Set/array; returns 0 for empty
   - `normalizeDashboardPart(part)` — folds `VA1`/`VA2`/`VA…` → `VA` for the Dashboard pie/bar
   - `parseOthers`, `stripParens`, `classOf`, `canonicalize` (helpers)
   - `extractUniquePlayers` — for the Player dropdown
@@ -126,7 +127,7 @@ Classes: `upper` (V1, V2, VA, VLA — violin/viola alias as one person) and `cel
 
 ### Calendar specifics
 
-Per-year stats column shows four numbers (Pieces, Unique Pieces, People played with, Playing Days) at `cellSize*2` through `cellSize*5`, with tooltips wired via `attachStatTooltip` (works on hover and tap). The stat values + tooltip copy live in `_yearStatDefs`, shared by both layouts (horizontal right-hand column and fullscreen below-grid rows) so they can't drift. The legend SVG is sized to exactly `10 * cellSize` wide and uses CSS `width: min(170px, 17%)` + `margin-left: min(40.5px, 4.05%)` so it tracks the calendar grid's first 10 cells across all viewport widths.
+Per-year stats column shows five numbers (Pieces, Unique Pieces, People played with, Playing Days, Longest Streak) at `cellSize*2` through `cellSize*6`, with tooltips wired via `attachStatTooltip` (works on hover and tap). The stat values + tooltip copy live in `_yearStatDefs`, shared by both layouts (horizontal right-hand column and fullscreen below-grid rows) so they can't drift. The legend SVG is sized to exactly `10 * cellSize` wide and uses CSS `width: min(170px, 17%)` + `margin-left: min(40.5px, 4.05%)` so it tracks the calendar grid's first 10 cells across all viewport widths.
 
 **Fullscreen (vertical) mode**: an expand button (network-graph style: `.network-fullscreen-btn` for looks + `.calendar-fullscreen-btn` for placement) toggles a lightbox where the grid renders transposed (`renderYearGroupsVertical`) — weeks run down, days across — with the cell size fitted so a whole year fills the viewport height (54 week rows: a leap year starting Saturday spans 54 Sunday-weeks). Year columns run chronologically left→right and the container opens scrolled to its right edge, so the current year is in view and panning left walks back in time. Legend + recent stats are omitted to give the grid every pixel. Entering also calls `requestFullscreen()` on `<html>` — NOT on `#calendar`, because browsers render only the fullscreen element's subtree and the tooltip div is a `<body>` child — with every failure path silently falling back to the fixed `100dvh` overlay CSS (older iOS, installed PWAs). Esc or the collapse button exits both native fullscreen and the lightbox; a gesture-exit from *native* fullscreen deliberately leaves the lightbox open (auto-closing on `fullscreenchange` closed everything whenever a browser bounced the request). A window `resize` listener re-fits the grid on rotation / chrome show-hide.
 
--- a/build.sh
+++ b/build.sh
@@ -159,10 +159,12 @@ else
         echo "Watching static assets with fswatch (PID $!)..."
 
         # Re-run tests on any change under src/ or test/. Dot reporter so each
-        # rerun is one compact line instead of 31 ✔'s.
+        # rerun is one compact line instead of 31 ✔'s. TZ matches the npm test
+        # script (and thus CI) so timezone-sensitive tests can't pass on save
+        # but fail in CI.
         fswatch -o --latency 0.5 src test | while read _; do
             echo "[$(date +%H:%M:%S)] JS change — re-running tests..."
-            node --test --test-reporter=dot test/*.mjs || true
+            TZ=America/New_York node --test --test-reporter=dot test/*.mjs || true
         done &
         BG_PIDS="$BG_PIDS $!"
         echo "Watching src/ + test/ for tests (PID $!)..."
--- a/package.json
+++ b/package.json
@@ -3,6 +3,6 @@
     "private": true,
     "type": "module",
     "scripts": {
-        "test": "node --test test/*.mjs"
+        "test": "TZ=America/New_York node --test test/*.mjs"
     }
 }
--- a/src/calendarComponent.js
+++ b/src/calendarComponent.js
@@ -799,8 +799,9 @@ export class CalendarComponent {
 
         // Full labels on desktop, dashboard-style short labels on mobile (the
         // two are toggled by a CSS media query, mirroring the dashboard tiles)
-        // so five metrics stay clear where there's room and still fit one line
-        // on a phone. The tap/hover tooltip spells out the full name either way.
+        // so five metrics stay clear where there's room and stay compact on a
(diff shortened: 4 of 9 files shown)


jsundram (commented):
## Review: max-streak metric

Adds a fifth aggregate metric — longest run of consecutive playing days — to `computeAggregateStats`, and surfaces it in the calendar's "Last 365 days" block, the per-year stats column (both layouts), the dashboard KPI tiles, and the Home ALL tab.

**The core logic is correct.** I verified the pieces that matter:

- `dayOrdinal` is genuinely DST-safe. Feeding local Y/M/D through `Date.UTC` yields exact multiples of 86400000, so consecutive calendar days always differ by exactly 1 — the `Math.floor` is a no-op but harmless, and it stays exact for pre-1970 negatives.
- Swapping `dayKey` → `dayOrdinal` doesn't change `daysPlayed`. Both are injective over local Y/M/D, so the distinct-day count is identical.
- `longestConsecutiveRun` handles empty (0), singleton (1), duplicates, and unsorted input. The `prev !== null` guard (rather than a falsy check) is the right call — ordinal 0 would break a truthiness test.
- The calendar's per-year streak comment is accurate: `d3.timeDay.range` produces a gap-free ascending day array and `d3.groups` preserves that order within each year, so scanning `value > 0` in order is valid.
- Horizontal layout has room for the fifth number. With `cellSize=17`, stats land at y=34…102 within a year group of height 170, and the grid bottom + weekly totals sit at 119/131 — no collision with the next year.
- The vertical layout's `bottomMargin = 16 + statDefs.length * statRowH + 8` derives from the array length, so the fifth row is accommodated automatically. Nice.

`npm test` passes (111/111).

---

### 1. The DST regression test is inert in CI — medium

`test/dataProcessor.test.mjs:420` is the one test guarding the change the PR is built around, and it only exercises that bug in a timezone that springs forward on 2026-03-08. I reimplemented the naive alternative (differencing local-midnight timestamps) and ran the same three dates through it:

```
TZ=UTC               naive streak = 3   (test asserts 3)  ← passes vacuously
TZ=Europe/Berlin     naive streak = 3   (test asserts 3)  ← passes vacuously
TZ=America/New_York  naive streak = 2   (test asserts 3)  ← actually catches it
```

Nothing in `package.json`, `build.sh`, or `.github/workflows/deploy.yml` sets `TZ`, and GitHub Actions runners default to UTC. So the guard passes on your Mac and is dead in CI — a future refactor back to timestamp-differencing would go green.

Cheapest fix is pinning the runner timezone in the test script:

```json
"test": "TZ=America/New_York node --test test/*.mjs"
```

That makes every test deterministic w.r.t. timezone, which is a win beyond this one case. If you'd rather not force a global TZ, the alternative is exporting `dayOrdinal` and unit-testing it directly against a fixed offset.

### 2. Stale "four" in comments — low

The PR updated the neighbouring lines but left the counts behind:

- `src/dashboardComponent.js:139` — "Same four metrics as the calendar's…" (the very next line now enumerates five)
- `src/dashboardComponent.js:188` — "the four tiles share one full-width line"
- `static/css/viz.css:862` — "all four tiles share one full-width line" (file untouched by this PR, but now wrong)

### 3. Five tiles in a `nowrap` mobile row may overflow — low/medium

`.stat-tiles` switches to `flex-wrap: nowrap` with `flex: 1 1 0` below 599px, and that row was dimensioned for four tiles. With five:

- 375px viewport → ~62px per tile → ~44px of content box after padding and border
- 320px viewport → ~51px per tile → ~33px of content box

`.stat-tile-value` renders at 18px/600 with no overflow handling — only `.stat-tile-label` gets `overflow: hidden; text-overflow: ellipsis`. A four-digit Pieces value is roughly 40px at that size, so it fits at 375px and spills past the tile border at 320px. The dashboard defaults to the `1Y` range where Pieces is probably three digits, but one tap on `All` gets you there.

Either let the row wrap on very narrow screens, or drop the mobile value to 16px:

```css
@media only screen and (max-width: 380px) {
    .stat-tiles { flex-wrap: wrap; }
}
```

Worth a look on a real phone with the `All` range selected before merging — I couldn't confirm the spill visually, only the arithmetic.

### 4. Three label vocabularies for the same five metrics — low

The same metrics now read differently in three places:

| | Calendar recent | Dashboard | ALL tab |
|---|---|---|---|
| unique pieces | `Unique` | `Unique pieces` / `Unique` | `Unique pieces` |
| unique people | `People` | `Unique people` / `People` | `Unique people` |
| days played | `Days` | `Days played` / `Days` | `Days played` |
| streak | `Streak` | `Max streak` / `Streak` | `Max streak` |

The commit message justifies the calendar shortening as "so all five fit on one line", but `.recent-stats-row` is already `display: flex; flex-wrap: wrap` — it would have wrapped cleanly rather than overflowed, so the shortening was a choice rather than a constraint. It costs some clarity: a bare `Days` sitting under a "Last 365 days" heading reads ambiguously, and `Unique` doesn't say unique-what.

The dashboard's long/short pair is the nicer pattern. Either adopt it in `renderRecentStats` (long labels, shortened under the mobile media query), or keep the long labels and let the row wrap to two lines.

### 5. The streak metric now has two implementations — low, optional

`_yearStatDefs` (`src/calendarComponent.js:361`) hand-rolls the run scan rather than calling the freshly-exported `longestConsecutiveRun`. The comment explains why (different input shape — a gap-free day array rather than ordinals), and I confirmed the two agree on the empty case, so there's no live drift. But the PR's stated goal for exporting the helper was a single definition, and this is one line away from using it:

```js
value: year => longestConsecutiveRun(
    yearQ.get(year).filter(d => d.value > 0).map(d => dayOrdinal(d.date))
)
```

That needs `dayOrdinal` exported. Reasonable to skip if you'd rather not widen the module's surface — the local scan is genuinely simpler.

### Minor

`longestConsecutiveRun` wraps its input in `new Set(days)` even when `computeAggregateStats` already passes a `Set`. Harmless (one extra allocation per call, at this data size irrelevant) and it makes the array path safe, so I'd leave it.

---

### Not an issue

- **Security** — no new input handling. The `title`/`desc` strings flowing into the tooltips' `.html()` are all static literals, no user data.
- **Performance** — one sort of at most (number of playing days) integers per aggregate call. Irrelevant even on the dashboard's re-render-everything-on-filter-change model.
- **Docs** — `CLAUDE.md` is updated accurately, including the `cellSize*2` through `cellSize*6` span, which matches `cellSize*(2 + i)` for i in 0…4.

### Verdict

The metric itself is correct and well-integrated, and the layout code was already parameterised well enough that adding a fifth stat mostly just worked. I'd fix the CI-inert DST test (#1) and the stale comments (#2) before merging, take a look at the mobile tile row on a narrow phone (#3), and treat #4 and #5 as follow-ups.

jsundram:
## Re-review of 9e4c052

All four items are addressed, and I verified each rather than reading the diff. One small gap left over from the TZ fix.

### 1. DST test — fixed, and I confirmed it now bites

The `TZ=America/New_York` pin works. Rather than trust the arithmetic, I mutation-tested it: reverted `dayOrdinal` to the naive form the comment warns about (`new Date(y, m, d).getTime() / 86400000`, i.e. dividing local-midnight timestamps instead of routing through `Date.UTC`) and ran the suite on the PR branch.

```
✖ maxStreak stays contiguous across a spring-forward DST boundary
ℹ tests 111   ℹ pass 110   ℹ fail 1
```

Exactly the one test, and nothing else. Before the pin that mutant passed 111/111. The guard is real now.

**One gap:** `build.sh:165` — the fswatch on-save rerun — calls the runner directly and so skips the npm script and the TZ pin:

```sh
node --test --test-reporter=dot test/*.mjs || true
```

So the dev watch loop and `npm test` / CI now run in different timezones, which is the kind of divergence that produces "passes on save, fails in CI." Prefixing it the same way is the one-line fix:

```sh
TZ=America/New_York node --test --test-reporter=dot test/*.mjs || true
```

`build.sh:87` and `:141` both go through `npm test`, so they're already covered.

### 2. Calendar labels — fixed, and the CSS-toggle approach is the better call

Rendering both spans and letting a media query pick one is sounder here than the dashboard's JS branch, because `renderRecentStats` has no resize re-render to hang a `s.mobile ? short : label` decision off. Checked the things that usually go wrong with this pattern:

- No phantom `gap` from the hidden span — `display: none` elements aren't flex items, so `.recent-stat`'s `gap: 4px` still yields one gap, not two.
- No duplicate announcement to screen readers; `display: none` drops the subtree from the a11y tree.
- The media query overrides `.recent-stat-label--short { display: none }` on source order at equal specificity. Correct as written.

I also checked whether the calendar's CSS *viewport* query can disagree with the dashboard's JS *element-width* breakpoint. In practice it can't: `#dashboard` is `box-sizing: border-box` under a 1000px max-width, so its rect width equals the viewport below that. The only divergence is the hidden-element fallback (`innerWidth - 40`), and `notifyShown()` re-renders at the real width once the view appears. Non-issue.

Nit: the new comment says the short labels "still fit one line on a phone." At 320px, `.calendar-top` leaves the stats roughly 245px after the legend's `min(170px, 17%)`, and five short stats at 10px want ~340px — so it wraps to two lines. Harmless, since `.recent-stats-row` is `flex-wrap: wrap` and always was, but the comment claims more than it delivers.

### 3. Mobile tiles — fixed, with about 5px of slack

Redid the arithmetic against the real containers (`#dashboard` is border-box, `max-width: 1000px`, `padding: 0 10px`, so a 320px viewport gives the row 300px):

| | before | after |
|---|---|---|
| tile content box | 37.2px | 41.2px |
| 4-digit value width | ~40px (18px font) | ~35.6px (16px font) |

So it moves from overflowing the tile border to clearing it. Still snug — five digits would spill — but that's 10,000+ pieces, so it's the right place to stop.

### 4. Stale "four" — all three fixed

`dashboardComponent.js:139`, `:188`, and `viz.css:865` are all corrected, and the CSS comment picked up a useful note about the calendar sharing the breakpoint.

### Left as-is

The DRY item (`_yearStatDefs` hand-rolling the run scan) and the redundant `new Set` wrap are both untouched, which is what I'd have chosen too — I'd flagged both as optional and neither can drift.

---

**Verdict:** ready to merge once `build.sh:165` picks up the TZ prefix. Everything else is verified.

jsundram:
## 08cf04e verified — merging

Both items from the re-review are addressed, and I checked the fix does what it claims rather than just reading the prefix:

- `TZ=America/New_York` in `build.sh:167` resolves to `GMT-0400` on 2026-03-09, so the watch loop really is running inside EDT and the spring-forward test is meaningful there, not just nominally pinned.
- The `--test-reporter=dot` flag stays ahead of the positional globs, so the watch loop keeps its compact output. Worth noting for the future: moving that flag after the globs makes Node silently ignore it — no error, just full spec output. That's why routing `build.sh` through `npm test --` wasn't an option and the `TZ` string had to be duplicated.
- The two `TZ` strings match exactly (`package.json`, `build.sh`).
- `bash -n build.sh` clean; full suite green at 111/111.

The comment correction on `renderRecentStats` is accurate — `.recent-stats-row` is `flex-wrap: wrap`, so wrapping is the designed behavior.

Squash-merging to match the repo's linear history. That triggers the Pages deploy to https://log.quartetroulette.com, which re-runs `npm test` before building.

PR #10, opened 2026-08-19
Add Unique Parts stat to summary cards and per-year calendar column

jsundram:
## What

Adds a sixth metric, **Unique Parts**, to the shared summary stat card (ALL tab, Dashboard KPI tiles, Calendar "Last 365 days" header) and to the calendar's per-year stats column.

**Definition**: distinct work + part combinations (`composer|title|part`). The same piece on V1 and later on V2 counts twice; repeats on the same part collapse to one; quintet/sextet second-viola (`VA2`) rows count separately from `VA` (`VA1` already folds into `VA` at `processRow` time). Rows without a part contribute to Pieces but not Unique Parts.

## How

- `computeAggregateStats` returns `uniqueParts`, built on new exported `workKey(d)` / `workPartKey(d)` helpers so the key semantics live in one place (the calendar's per-year accumulation uses the same helpers — the two can't drift).
- New shared def in `statDefs.js` after Unique pieces (label "Unique parts", mobile short label "Parts"); the VA2 tooltip explainer is single-sourced as `UNIQUE_PARTS_DESC`, reused by `_yearStatDefs`.
- Calendar per-year column: new `yearParts` map threaded through both layouts. No current-year projection — the count saturates like Unique pieces, so a linear extrapolation would overshoot.
- Layout verified count-agnostic: the three card renderers iterate the def array; mobile CSS gives tiles equal flex shares; the horizontal calendar's sixth stat at `cellSize*7` stays inside the year band; the fullscreen layout sizes its bottom margin from `statDefs.length`.

## Testing

- New unit tests: VA/VA2 distinctness, same-part collapse, cross-composer keys, partless rows; updated def-order/zero-stats fixtures.
- e2e smoke updated (KPI tile count 5 → 6).
- Verified visually at a 390px mobile viewport against real data (all three views).

https://claude.ai/code/session_01PWxB5xbG9jDkHpNpciKnkv


static/css/viz.css
```diff
@@ -893,7 +893,7 @@ body.calendar-fullscreen-open {
 
     /* Tighter padding + smaller value than the four-tile layout used, so a
        four-digit value (the All range) still clears the tile border when
-       five share a 320px-wide row. */
+       six share a 320px-wide row. */
```
jsundram: **[1] Mobile six-tile row: a four-digit value now overflows its tile border.** The comments were updated five→six but the tuning values (`gap: 6px`, `padding: 6px`, `font-size: 16px`, `flex-wrap: nowrap`) were not. On a 320px viewport: `#dashboard`'s 300px content minus five 6px gaps → six 45px tiles; after 2px border + 12px padding the content box is ~31px, but a 4-digit value (the All range, live data) at 600-weight 16px sans is ~35px. `.stat-tile-value` has no overflow handling (only the label has ellipsis) and the row is `nowrap`, so the number paints past the rounded tile border. With five tiles (~41px content) it fit — this comment's claim is now false. Fix: retune (smaller value font / tighter padding at this width) or give `.stat-tile-value` overflow handling.

src/dataProcessor.js
```diff
@@ -245,29 +245,56 @@ export function formatStreakStart({ count, start }) {
...
+}
+
+// Unique parts key on the raw part value, so a quintet played on VA and
+// again on VA2 counts as two parts of the same work (VA1 is already folded
+// into VA at processRow time).
+/**
+ * @param {Row} d
+ * @returns {string|null} key for unique-part counting, null if untitled or partless
+ */
+export function workPartKey(d) {
```
jsundram, line 265: **[2] `workPartKey` counts VA2 as a distinct part while every part filter folds VA2 into VA — the new stat can contradict the active single-part filter.** With the VA part button selected (Home/ALL via `filterEngine.partMatches` `startsWith('VA')`) or VA selected on the dashboard (`normalizeDashboardPart` fold), VA2 rows pass the filter, but this key stays raw — so one quintet played once on VA and once on VA2 shows *Unique pieces: 1 / Unique parts: 2* under a single selected part, and disagrees with the part chart directly below showing one VA slice. Reachable with current data (VA2 rows exist). `UNIQUE_PARTS_DESC` documents the split as intentional, but nothing explains parts > pieces under a one-part filter; consider folding to the filter's granularity when a part filter is active, or noting the disagreement in the tooltip.

src/dataProcessor.js
```diff
@@ -245,29 +245,56 @@ export function formatStreakStart({ count, start }) {
...
+    const key = workKey(d);
+    return key && d.part ? `${key}|${d.part}` : null;
```
jsundram, line 267: **[3] Untrimmed `part` is now identity-bearing — stray whitespace silently inflates the headline stat.** *(Anchored here; root is `processRow`, line 636, not in this diff.)* `processRow` trims every sheet field defensively except `part` (`"part": r["Which Part"] == "VA1" ? "VA" : r["Which Part"]`), and `fillForward` never touches part. Before this diff an untrimmed `'V1 '` caused only a filter miss; now it mints a second Set key (`Haydn|17#1|V1` vs `Haydn|17#1|V1 `) and a wrong number in three headline surfaces (ALL tab, dashboard KPI, calendar per-year column) with no visible signal — and `' VA1'` also defeats the exact `== "VA1"` fold. One-word fix: trim in `processRow` before the comparison. (Current data snapshot is clean; this is latent dirty-data hardening.)

src/calendarComponent.js
```diff
@@ -137,15 +137,22 @@ export class CalendarComponent {
 
         // Per-year count of unique pieces (composer + work title).
         const yearUnique = new Map(years.map(([y]) => [y, new Set()]));
+        // Per-year count of unique work + part combinations (workPartKey:
+        // raw part value, so quintet VA and VA2 rows count separately).
+        const yearParts = new Map(years.map(([y]) => [y, new Set()]));
```
jsundram, line 142: **[4] The per-year stats column is a hand-rolled parallel `computeAggregateStats`, and this diff deepens it.** Adding the stat threaded a third parallel Map through six coordinated sites (Map construction, accumulation loop, guard, config literal, two destructurings, `_yearStatDefs` signature). Grouping each year's rows and calling the existing `computeAggregateStats` would return all six numbers with no plumbing. The two paths already diverge subtly with nothing cross-checking them — e.g. `_yearStreak` hands `formatStreakStart` a local-midnight `d3.timeDay` Date where `computeAggregateStats`' convention stores the local day in UTC fields (off-by-one streak-start display at positive UTC offsets) — so a rule change applied to one path silently misses the other.

e2e/smoke.spec.js
```diff
@@ -64,7 +64,7 @@ test('calendar view renders the year grid', async ({ page }) => {
 test('dashboard view renders KPI tiles and charts', async ({ page }) => {
     await page.evaluate(() => { window.location.hash = '#dashboard'; });
     await expect(page.locator('#dashboard')).toBeVisible();
-    await expect(page.locator('#dashboardStats .stat-tile')).toHaveCount(5);
+    await expect(page.locator('#dashboardStats .stat-tile')).toHaveCount(6);
```
jsundram, line 73: **[5] Fixture is degenerate for the new stat, and the calendar's `yearParts` path has no test at all.** The four in-window fixture rows have four distinct works AND four distinct (work, part) pairs, so `pieces == uniquePieces == uniqueParts == 4` — a tile mis-wired to `agg.uniquePieces` renders identically here. The def-to-field wiring is covered by `presenter.test.mjs` (17 vs 21 fixture), but nothing constructs `CalendarComponent` in test/ or e2e/, so the `yearParts` accumulation (calendarComponent.js:140–157) could regress with the suite green. One extra fixture row repeating a work on a different part would make the stat independently observable end-to-end.

src/calendarComponent.js
```diff
@@ -363,10 +371,10 @@ export class CalendarComponent {
             .text(d => d > 0 ? d : "");
     }
 
-    // The five per-year stats (value + tooltip content), shared by the
+    // The six per-year stats (value + tooltip content), shared by the
     // horizontal layout's right-hand column and the vertical layout's
     // below-grid rows so the numbers and explanations can't drift apart.
-    _yearStatDefs(yearQ, yearUnique, yearPeople) {
+    _yearStatDefs(yearQ, yearUnique, yearParts, yearPeople) {
```
jsundram: **[6] Four positional params, three of which are interchangeable `Map<number, Set>` — a transposition would be silent.** This file has no `// @ts-check` and no test covers `_yearStatDefs` or either layout, so swapping `yearParts`/`yearPeople` at either call site (lines 342, 462) would show wrong numbers under the wrong tooltip in one layout with no error. Cheap fix: pass the existing `config` object (it already carries all three maps under these names) and destructure inside, so the signature never grows another positional arg.

jsundram (commented):
Code review of the Unique Parts stat (high-effort recall pass): 9 findings, ranked most-severe first in each comment's **[N]** prefix. Findings [1]–[3] are correctness issues, [4]–[9] are cleanup/robustness/test-coverage. Three comments are anchored to the nearest changed line because their exact location isn't in this diff: [3] concerns `processRow` at src/dataProcessor.js:636, [7] concerns `bottomMargin` at src/calendarComponent.js:480, [8] concerns the stat placement at src/calendarComponent.js:349.

src/calendarComponent.js
```diff
@@ -443,9 +457,9 @@ export class CalendarComponent {
     // fits the viewport height on a portrait phone. One column per year,
     // most recent leftmost; the container pans horizontally across years.
     renderYearGroupsVertical(container, years, config) {
-        const { timeWeek, formatDay, formatMonth, formatDate, countDay, color, sessions, yearUnique, yearPeople } = config;
+        const { timeWeek, formatDay, formatMonth, formatDate, countDay, color, sessions, yearUnique, yearParts, yearPeople } = config;
         const yearQ = new Map(years);
-        const statDefs = this._yearStatDefs(yearQ, yearUnique, yearPeople);
+        const statDefs = this._yearStatDefs(yearQ, yearUnique, yearParts, yearPeople);
```
jsundram: **[7] Sixth stat row grows fullscreen `bottomMargin` 94→108, pushing availH ≈ 514–527px viewports from exact fit into overflow.** *(Anchored here; the math is at lines 477–485, not in this diff.)* At availH = 514: pre-diff `cell = floor((514-136)/54) = 7` exactly, svgH 514 fits; post-diff `floor((514-150)/54) = 6` clamps up to the min of 7, svgH = 528 > availH, so the streak row falls below the fold of a lightbox whose premise is “a whole year fits the viewport height” — and all shorter viewports overflow 14px more than before. Mitigated by the `overflow: auto` fallback on `#calendar.fullscreen`; affects the CSS-overlay fallback path (short landscape phones, small desktop windows), not native fullscreen. Minor, but worth knowing the fit band moved.

src/calendarComponent.js
```diff
@@ -331,7 +339,7 @@ export class CalendarComponent {
         // Per-year stats (shifted right to make room for day-of-week totals).
         // One stacked bare number per stat; the tooltip explains which is which.
         const yearQ = new Map(years);
-        this._yearStatDefs(yearQ, yearUnique, yearPeople).forEach((def, i) => {
+        this._yearStatDefs(yearQ, yearUnique, yearParts, yearPeople).forEach((def, i) => {
```
jsundram: **[8] Horizontal stat column has no bound derived from `statDefs.length` — this diff consumed the second-to-last slot on luck.** *(Anchored here; placement is `cellSize*(2 + i)` at line 349, not in this diff.)* The per-year band is a fixed `CALENDAR_CONFIG.height = 17*10` (src/config.js:82); the sixth stat lands at y=119 and a seventh at y=136 still fits, but an eighth (y=153) silently spills into the next year's band and clips past the SVG height — no assertion, no test. The vertical layout already derives its `bottomMargin` from `statDefs.length` (line 479); derive or assert the horizontal band the same way.

src/calendarComponent.js
```diff
@@ -388,6 +396,12 @@ export class CalendarComponent {
                 title: year => `Unique pieces played in ${year}`,
                 desc: () => "Distinct works (composer + title) logged this year. Partial-movement entries don't count, so repeats of the same piece collapse to one."
             },
+            {
+                label: "parts",
+                value: year => yearParts.get(year)?.size ?? 0,
+                title: year => `Unique parts played in ${year}`,
+                desc: () => UNIQUE_PARTS_DESC
```
jsundram: **[9] (nit) The per-year parts explainer is the only one of the six that isn't year-scoped.** Its five siblings say “logged this year” / “played in {year}” in the desc body, while the shared `UNIQUE_PARTS_DESC` is window-agnostic (it serves the ALL tab / dashboard / Last-365 too). The tooltip *title* does carry the year, so impact is small — but an editor sweeping the calendar tooltip copy will miss the one desc that lives in `statDefs.js`.

jsundram:
Review findings addressed in 19e8013 — disposition per finding:

- **[1] Fixed.** Mobile tile padding tightened to `6px 4px` and `.stat-tile-value` capped at `min(16px, 4.5vw)`, so the value scales with the six-across row below ~390px instead of holding 16px into a tile that can't fit it. Verified with live data at a 320px viewport on the All range (Pieces = 3219, four digits): 0px overflow on all six tiles, screenshot-checked.
- **[2] Addressed via tooltip.** VA2-distinct is the intended counting rule, so the semantics stand; the shared desc now states the asymmetry explicitly ("…count separately from VA, while the VA part filter includes them — so a VA-filtered view can show more unique parts than unique pieces"). Folding to filter granularity would make the stat's meaning depend on which control produced the slice — happy to revisit if the tooltip proves insufficient.
- **[3] Fixed.** `processRow` now trims part before the `VA1` fold; unit tests cover `' VA1 '` → `VA`, `'V1 '` → `V1`, and whitespace-only → `''`.
- **[4] Tracked in #11**, which was filed for exactly this refactor before the review ran; the `_yearStreak` local-vs-UTC streak-start divergence is now noted there as further evidence.
- **[5] Fixed.** Fixture adds `20#2` on a new part plus a repeat of that (work, part), making Pieces (6) / Unique pieces (4) / Unique parts (5) pairwise distinct; each of the first three KPI tiles now asserts its own value. The `yearParts` accumulation itself remains uncovered directly (no `CalendarComponent` construction in tests), but it now shares `workKey`/`workPartKey` with the unit-tested aggregate path.
- **[6] Fixed.** `_yearStatDefs(yearQ, config)` — the three `Map<year, Set>`s are destructured by name inside, so a call-site transposition is no longer expressible.
- **[7] Acknowledged, no change.** The fullscreen fit band did move by one 14px stat row; the `overflow: auto` fallback covers the affected CSS-overlay path. Will fold into #11's rework if the band math changes anyway.
- **[8] Fixed.** `renderYearGroups` now `console.assert`s `statDefs.length` against the band's derived capacity (`floor(height/cellSize) - 3` = 7), so an eighth stat fails loudly in dev instead of spilling into the next year's band.
- **[9] Fixed.** `UNIQUE_PARTS_DESC` became `uniquePartsDesc(scope)`; the calendar passes `' this year'` so its copy is year-scoped like its five siblings while staying single-sourced.

All checks pass post-fix: 279 unit tests, lint, typecheck, and the 3 Playwright smokes.

https://claude.ai/code/session_01PWxB5xbG9jDkHpNpciKnkv

PR #17, opened 2026-08-23
keep pull-to-refresh out of gestures that belong to a scroller

jsundram:
## The bug

Open the Player filter on Home in the installed PWA, with enough names that the list scrolls. Swiping up through the list works; swiping back down freezes the list and shows the refresh spinner instead.

`.player-dropdown` is a `max-height: 300px; overflow-y: auto` list, and `#playerSelect` sits directly under the `<h1>` (`index.html:129`), so `window.scrollY` is 0 the whole time it's open. `PullToRefresh` gated on exactly that and never looked at `e.target`:

```js
_onStart(e) {
    if (this._scrollTop() > 0) { this.startY = null; return; }  // page scroll only
    this.startY = e.touches[0].clientY;
}
_onMove(e) {
    const delta = e.touches[0].clientY - this.startY;
    if (delta <= 0) return;
    e.preventDefault();   // ← kills the dropdown's own scroll
```

The `delta <= 0` early-out is what makes it directional: swiping up returns before `preventDefault()` and scrolls natively; swiping down gets claimed as a pull. One direction working and the other not is why it reads as the app not knowing which gesture is which.

## The fix

A pure, exported `startsInScroller(target, stopAt, styleOf)` walks `e.target`'s ancestors to `<body>` and returns true if any is declared scrollable (`auto`/`scroll`) on **either** axis and actually overflows on that axis. `_onStart` gates on it alongside the page-scroll check:

```js
if (this._scrollTop() > 0 || this._startedInScroller(e.target)) {
    this._reset();
    return;
}
```

Structural rather than a class allowlist, mirroring the tooltip's ownership walk — so the scrollable tooltip (`static/css/viz.css:575-581`) and the fullscreen lightboxes are covered too, and a future scrollable panel is covered for free.

Both axes matter because PTR only ever sees the vertical component of a gesture. Without the Y check, a downward drag inside the dropdown is `preventDefault()`'d into a pull and the list can't scroll back up. Without the X check, a *sideways* swipe with a few pixels of drift is claimed the same way — which hits `#tabs` and `.date-filter-container` (both `overflow-x: auto`, both sitting in the top-of-page band where `scrollY === 0`) and the fullscreen calendar, whose year columns pan while its height is fitted to the viewport, so it has no vertical overflow to detect and `body.calendar-fullscreen-open` pins `scrollY` past the `_scrollTop()` guard.

The `_reset()` (rather than clearing `startY` alone) matters because the new half of the condition can fire *mid-pull* — a second finger landing inside a scroller — and `_onEnd()` early-returns on a null `startY`, which would strand the half-pulled indicator on screen.

## Verification

`test/pullToRefresh.test.mjs` — 13 cases over the walk, using plain objects with `parentElement` chains as stand-in nodes (same pattern as `tooltip.test.mjs`): the dropdown, a horizontal pan, the fullscreen calendar's viewport-fitted grid, a wrapper declared scrollable on both axes that overflows on neither, `overflow: hidden`, the boundary node, a text-node target with no computed style, a detached chain — plus one that seeds a half-pulled state and asserts `_onStart`'s bail clears it.

Behavior checked in a touch-enabled Chromium with `matchMedia('(display-mode: standalone)')` stubbed true, driving gestures through CDP touch events with `window.scrollY` forced to 0:

| gesture | before | after |
| --- | --- | --- |
| drag down in a 46-name dropdown | `pulling: true`, `listScrollTop: 120` (frozen) | `pulling: false`, `listScrollTop: 45` |
| sideways swipe across `#tabs` with downward drift | `pulling: true`, `scrollLeft: 200` (frozen) | `pulling: false`, `scrollLeft: 106` |
| drag down on the page background | pulls | pulls |

Full suite green (302 pass), lint and typecheck clean.

## Trade-off, stated rather than implied

A genuinely *vertical* pull that starts on `#tabs`, `.date-filter-container`, or a table wrapper no longer refreshes once that element actually overflows horizontally. The page around them stays pullable. Not stealing the pan is worth more than the target area.

I considered axis-locking in `_onMove` instead (classify by dominant axis on the first move, keeping those strips vertically pullable) and rejected it: deferring the `preventDefault()` past the first `touchmove` risks iOS handing the gesture to the compositor and ignoring the later call, which would break PTR on the one platform it exists for. The structural test decided at `touchstart` has no such timing hazard.

Also left alone: a drag starting on the collapsed `.player-select-trigger` row still counts as a pull, and the check doesn't consult a scroller's own `scrollTop` — a list already at its top hands the gesture to nothing rather than to PTR, which seems right inside an open overlay.


src/pullToRefresh.js
```diff
@@ -9,6 +9,31 @@ const THRESHOLD = 80;  // px of pulled distance past which release triggers refr
...
+// target and getComputedStyle would throw on them. An element that only
+// scrolls horizontally (the data tables' overflow-x wrappers, which the
+// cascade computes overflow-y:auto on) fails the scrollHeight check and is
+// walked past. Pure — exported for tests.
+export function startsInScroller(target, stopAt, styleOf) {
+    for (let el = target; el && el !== stopAt; el = el.parentElement) {
+        const style = styleOf(el);
+        if (!style) continue;
+        const scrollable = style.overflowY === 'auto' || style.overflowY === 'scroll';
+        if (scrollable && el.scrollHeight - el.clientHeight > 1) return true;
```
jsundram: This only sees **vertical** overflow, so a horizontal-only scroller gets walked past and PTR eats its pan.

Concretely, `#calendar.fullscreen` (`static/css/viz.css:678-680`):

```css
/* The year columns pan horizontally; vertical scroll is a fallback for
   very short viewports where the min cell size doesn't quite fit. */
overflow: auto;
```

Its height is deliberately fitted to the viewport, so `scrollHeight - clientHeight` is ~0 and this check fails — even though the element is very much a scroller, just on the other axis. And `body.calendar-fullscreen-open { overflow: hidden }` pins `scrollY` at 0, so the `_scrollTop()` guard in `_onStart` doesn't catch it either. Net effect on a standalone-PWA install: a horizontal year-pan with a few px of downward drift hits `delta > 0` in `_onMove` → `preventDefault()` → the pan freezes, the spinner appears over the lightbox, and past the 80px threshold it fires a full refresh. Same shape for the `#tabs` strip and `.table-wrapper` (the comment above calls those out as intentionally still pullable, which is right for a *vertical* pull, but their horizontal pans have the same problem).

This also makes the new CLAUDE.md sentence inaccurate where it lists "the fullscreen lightboxes" as covered by the walk.

Suggested fix — treat either axis as a scroller:

```js
const scrollsY = (style.overflowY === 'auto' || style.overflowY === 'scroll')
    && el.scrollHeight - el.clientHeight > 1;
const scrollsX = (style.overflowX === 'auto' || style.overflowX === 'scroll')
    && el.scrollWidth - el.clientWidth > 1;
if (scrollsY || scrollsX) return true;
```

That keeps the structural spirit of the test, and the tables' `overflow-x` wrappers stay excluded from *vertical* pulls only when they genuinely don't overflow horizontally — which is the honest answer, since when they do overflow horizontally the user is panning them, not pulling the page.

---

src/pullToRefresh.js
```diff
@@ -44,9 +69,14 @@ export class PullToRefresh {
...
+    _startedInScroller(target) {
+        return startsInScroller(target, document.body,
+            (el) => (el.nodeType === 1 ? window.getComputedStyle(el) : null));
+    }
+
     _onStart(e) {
         if (this.refreshing) return;
-        if (this._scrollTop() > 0) {
+        if (this._scrollTop() > 0 || this._startedInScroller(e.target)) {
             this.startY = null;
```
jsundram: The new half of this condition can fire *mid-pull*, and the bail body isn't written for that: it sets `startY = null` but leaves `currentPull` and the indicator's `pulling` class untouched, while `_onEnd()` early-returns on `startY === null`.

So: finger down at the top of the page, pull halfway, then a second finger lands inside the open Player dropdown → `_onStart` bails → the half-pulled spinner is stranded on screen until some later unrelated gesture completes a full `_onStart`/`_onEnd` cycle.

The pre-existing `_scrollTop() > 0` half couldn't do this (scrollY is 0 for the whole duration of a pull, which is the precondition for being in one), so this is new with the change.

```suggestion
        if (this._scrollTop() > 0 || this._startedInScroller(e.target)) {
            this._reset();
```

`_reset()` already clears `startY`, `currentPull`, and the indicator, so it's a strict superset of the current behaviour. Cosmetic and rare, but cheap to close.

---

jsundram (commented):
Core fix is sound. The ancestry walk terminates correctly (`continue` in a `for` still runs the update expression), handles text-node touch targets via the `styleOf` → `null` path, stops exclusive of `<body>`, and injecting `styleOf` keeps it testable — the 11 new unit tests pass. I checked the comment's CSS claims against `static/css/viz.css` and `src/tableComponent.js`: `.table-wrapper`, `#tabs`, `.squares-container` and `.date-filter-container` all compute to auto height, so they do stay pullable as described.

Two findings inline, neither blocking the dropdown bug this sets out to fix:

1. **`startsInScroller` misses horizontal-only scrollers** — real regression on the fullscreen calendar in standalone PWA mode, where a horizontal year-pan can get `preventDefault()`ed into a refresh.
2. **The new `_onStart` bail can strand the indicator mid-pull** — cosmetic, needs a second finger to hit.

---

src/pullToRefresh.js
jsundram: Confirmed and fixed in 84fcfec — took the suggestion, checking both axes.

Reproduced it first. In a standalone-PWA Chromium, a sideways swipe across `#tabs` (`scrollWidth` 938, `clientWidth` 520) with ~1.5px/step of downward drift, on the previous head:

```
tab-strip pan: { pulling: true, scrollLeft: 200 }   ← pinned at its start value
```

After the fix the strip pans and PTR stays out:

```
tab-strip pan: { pulling: false, scrollLeft: 106 }
```

Worth noting the tab strip and `.date-filter-container` (`static/css/viz.css:347`) are the sharper cases than the fullscreen calendar — both are `overflow-x: auto` and both sit in the top-of-page band where `scrollY === 0`, so they were in the PTR zone on every visit, not just inside a lightbox.

One trade-off I've made explicit in the code comment and in CLAUDE.md rather than silently: this also means a genuinely *vertical* pull that starts on one of those strips no longer refreshes once it actually overflows horizontally. I considered axis-locking in `_onMove` instead (classify by dominant axis on first move, so the tables stay vertically pullable), but rejected it — deferring the `preventDefault()` past the first `touchmove` risks iOS handing the gesture to the compositor and ignoring the later call, which would break PTR on the one platform it exists for. The structural test decided at `touchstart` has no such timing hazard.

CLAUDE.md now says "either axis", names the strips, and explains why the fullscreen calendar slips both the Y check and the `_scrollTop()` guard.

---

src/pullToRefresh.js
jsundram: Right on all counts — applied in 84fcfec. `_reset()` is a strict superset, and the analysis of why the pre-existing `_scrollTop() > 0` half couldn't strand the indicator (scrollY is 0 for the whole duration of a pull, which is the precondition for being in one) is what makes this new with the change.

Added a unit test that pins it: seed `startY`/`currentPull` and a `pulling` indicator, stub `_scrollTop`/`_startedInScroller`, call `_onStart`, assert `currentPull === 0`, the class gone, and both custom properties back to zero.

---

jsundram:
Re-reviewed at `84fcfec`. Both findings are addressed, and the fix went further than the report on the first one — I'd only flagged `#calendar.fullscreen` and guessed at `#tabs`; the commit correctly found `.date-filter-container` too.

Checked on the new head: 302 tests pass, `npm run lint` and `npm run typecheck` are both silent. The `scrolls(overflow, scrollSize, clientSize)` helper applied per-axis keeps declared-scrollable-but-not-overflowing passing through, and the stale "horizontal-only scrollers stay pullable" test was replaced rather than left contradicting the new behaviour. The `_reset()` bail can't clobber a live refresh either, since `_onStart` already early-returns on `this.refreshing`.

**One follow-up thought, not a blocker.** The commit message is upfront that the fix costs pull area — *"a genuinely vertical pull starting on a horizontally-overflowing strip no longer refreshes"* — and that's the right priority ordering. But the surface is a bit larger than that sentence suggests. The four `overflow-x: auto` elements in `static/css/viz.css` are:

| line | selector |
| --- | --- |
| 347 | `.date-filter-container` |
| 491 | `#tabs` |
| 1177 | `#dashboardMusicianNetworkMatrix` |
| 1244 | `.squares-container` |

On a phone at `scrollY === 0` on Home, three of those four are most of what's above the fold, which leaves the `<h1>` and the part-button row as the pullable target. Worth a try on a real device to see whether it still feels reachable.

If it doesn't, the sharper fix is to decide by **gesture direction** rather than by element: record `startX` alongside `startY` on touchstart, note that the touch began in an X-scroller, and on the first `_onMove` bail only when `|dx| > |dy|`. That keeps both the pan and the vertical pull instead of trading one for the other. It's more state and more to test, so it seems like a follow-up rather than a reason to hold this PR — the current behaviour is strictly better than the bug it replaces.

---

jsundram:
Withdrawing the axis-locking suggestion from my previous comment — I'd read the commit but not the review-thread replies, where it had already been considered and rejected for a better reason than I had for proposing it:

> deferring the `preventDefault()` past the first `touchmove` risks iOS handing the gesture to the compositor and ignoring the later call, which would break PTR on the one platform it exists for

That's right, and it's decisive. A `passive: false` `touchmove` handler that lets the first move through can lose the gesture to the compositor, after which `preventDefault()` is a no-op — so classifying by dominant axis on first move trades a bounded loss of pull area for an unbounded risk of PTR not working at all. Deciding structurally at `touchstart` is the correct call.

The `overflow-x` selector table in that comment still stands as a note on how much above-the-fold area the tradeoff covers, but treat it as informational, not as an argument for changing the approach. No action needed here.

---

PR #18, opened 2026-08-27
Audit tooling for player-name and ensemble data quality

jsundram:
Three audits and one pipeline change, from a session spent chasing an inaccurate unique-player count. The count moved 359 → 342; most of that came from fixing the source sheet, and these are the tools that found what to fix.

## Pipeline change

**Instrument annotations on player slots are now honored.** The three player slots model a string quartet, so class and part came from position: slots 1–2 upper, slot 3 cello. Ensembles with no seat in that layout — piano trios and quartets above all — push people into whichever column is free, and a pianist landing in an upper slot was counted as a violinist. `stripParens` already *parsed* an `(instrument)` suffix and threw it away; the raw sheet had six that never survived.

`instrumentFromSlot()` now keeps it, and it overrides position — `classOf(annotation) ?? SLOT_CLASS[i]` when aliasing, `partFromInstrument(annotation)` over `SLOT_TO_PART` for the part breakdown. Unannotated slots behave exactly as before, so this is a no-op for every row without one.

`classOf` and `partFromInstrument` also share `CELLO_INSTRUMENT` / `VIOLA_INSTRUMENT` patterns instead of each carrying its own prefix checks. Only terse codes worked before: a cellist written `(cello)` — the most natural spelling, and the first thing a new logger would type — classified as `upper` **and** bucketed as `OTHER`, vanishing from the VC stats. The `(?![a-z])` guards stop `c` matching `clarinet` and `va` matching `violin`. Unnumbered violin stays `OTHER` deliberately: it says nothing about which violin seat.

## New audits

**`report_ambiguity()` in `audit_aliases.py`** — the complement to the existing variant grouping. Not "which short forms belong in `PLAYER_ALIASES`" but which must not, because an alias maps one name to one person and the sheet may hold several who share it. Three hazards: a bare first name several full names could match (unfixable by any alias — the rows need editing); an alias keyed on such a name, silently resolving every future bare entry; and an alias whose canonical name appears nowhere in the sheet.

It also warns when run against `archive/data.csv`. That export is written *after* `normalizePlayerNames`, so a working alias looks dead there — its bare form is missing precisely because the alias did its job. Acting on that reasoning during this session pruned four live aliases and un-normalized 66 rows, and the damage only surfaced two fetches later.

**`audit_ensembles.py`** — row completeness rather than names. Work titles imply a headcount, and rows logging fewer people than the work needs get reported, split by whether the ensemble was stated or the quartet default assumed. It separately lists piano works where nobody carries a keyboard annotation: headcount fine, pianist sitting in a string seat.

Titles are often catalogue numbers (`K478`, `1#1`) with the ensemble named only in the comment, so Comments are read too — but only when an instrument sits immediately before the ensemble word. Comments are prose and this is a music log: "more piano the second time" is a dynamic, "quintets were averted briefly" is a joke.

**`audit_fillforward.py`** — `fillForward` carries `player1/2/3` and `location` into the next row of a session but not `others`. A session logged the natural way (spell the group out once, leave the slots blank after) keeps the quartet and silently drops the second violist, the extra cellist, the pianist. It found 45 such rows across 26 sessions.

Only rows with every player slot blank are reported, since those unambiguously mean "same group as before" — a row re-typing some players may be dropping the extra person on purpose. That distinction is why this is a report and not an extra column in `fillForward`, which would touch 199 rows instead of 45.

## Running them

`npm run audit` runs all three against a freshly fetched sheet and prints a summary; `npm run audit -- --no-fetch` uses whatever is already in `archive/`.

The wrapper exists mainly to encode which file each audit needs, because getting that wrong caused the worst bug of this session. `audit_aliases` and `audit_ensembles` read the processed export — post-`fillForward`, so every row lists its full group, where on the raw sheet every continuation row would look under-logged. `audit_fillforward` reads the raw sheet by necessity, since it looks for exactly the blank slots `fillForward` erases. The summary's dangling-alias count comes from a second pass over raw, because on a canonicalized export every canonical name is present by construction and the count is always 0.

The full output runs to hundreds of lines, most of it groups that are fine, so the summary marks which findings decay. Bare ambiguous names and under-logged rows need memory and get harder to answer every month; a dropped `Others?` player is recoverable from the row above it whenever you get to it. That distinction is what lets the mechanical work batch up while the memory-dependent work stays fresh.

## Notes

- `md/howto.md` gains a section on logging non-quartet ensembles, including that `Others?` does not carry forward — which decides whether to annotate a slot or use `Others?`.
- Tests 283 → 295. Lint and typecheck clean.
- No real names appear in any tracked file; the tracked diff was checked against the live alias table.


scripts/audit_ensembles.py
```diff
@@ -0,0 +1,183 @@
...
+    first comma."""
+    m = SLOT_ANNOTATION_RE.match((value or "").strip())
+    if not m:
+        return None
+    inside = m.group(2)
+    comma = inside.find(",")
+    return (inside[:comma] if comma >= 0 else inside).strip() or None
+
+
+def has_keyboard_annotation(row: dict) -> bool:
```
jsundram, line 151: **The slot half of this check can never fire against the default input.**

`archive/data.csv` is written by `scripts/fetch_processed.mjs` → `serializeRows` → `rowToFields`, which emits `d.player1..3` *after* `normalizePlayerNames` has run `stripParens` on them. The `(instrument)` suffix is gone from the player columns in that file — only `Others?` is exported verbatim.

Verified against the real pipeline:

```
input  Player 3 = "Alice Hart (p)"
export Player 3 = "Alice Hart"        # annotation stripped
```

So `slot_instrument(row["Player 3"])` always returns `None` here, and a row where the pianist *is* correctly annotated in a player slot is still reported under **UNANNOTATED PIANO WORKS** — with a fix instruction (`annotate the slot in place`) that will never clear it. `audit_all.sh` surfaces the same inflated number as `piano works with nobody marked at the keyboard`.

Either read this check from `archive/data-raw.csv` (like `audit_fillforward.py` does), or have `rowToFields` preserve the slot annotation.

src/dataProcessor.js
```diff
@@ -80,6 +99,22 @@ export function stripParens(name) {
...
 
+// Pull the "(instrument)" suffix off a player-slot value, mirroring the
+// Others?-entry shape: inside the parens the first comma separates the
+// instrument code from a free-form comment, so "Alice Hart (vc, doubling)"
+// yields "vc". Returns null when the slot carries no annotation.
+/**
+ * @param {string|null|undefined} name
+ * @returns {string|null}
+ */
+export function instrumentFromSlot(name) {
```
jsundram, line 143: **Any trailing parenthetical is now read as an instrument, not just instrument annotations.**

`stripParens` deliberately discarded whatever was in the trailing parens; `instrumentFromSlot` now feeds it to `classOf` and `partFromInstrument`, and `classOf` never returns `null` for a non-empty string — so the `?? SLOT_CLASS[i]` fallback below cannot rescue a non-instrument note.

Concrete: `Player 3 = "Jo (sub)"`.
- Before: `classOf(null)` → fall back to `SLOT_CLASS[2] === 'cello'` → canonicalizes to `Jo Beta`.
- After: `classOf('sub') === 'upper'` → canonicalizes to `Jo Alpha` — the wrong person, silently.

And in `computePartBreakdownPerMusician`, `partFromInstrument('sub')` → `OTHER`, so that cellist drops out of the VC column entirely. Same for `(guest)`, `(sub)`, `(first time)`, `(Bob's teacher)`.

Consider only honoring the annotation when it actually classifies — e.g. return `null` unless `classOf`/`partFromInstrument` recognizes it, so unrecognized parentheticals keep the old positional behavior.

md/howto.md
```diff
@@ -46,7 +46,53 @@
...
+
+**Say what someone played with `(instrument)`.** Anyone whose instrument isn't
+the one their seat implies needs an annotation — most often the pianist, but
+also a cellist you had to seat in a violin field because the pianist took the
+cello field. The annotation wins over the seat, so this is correct even though
+the pianist is in the cello field:
+
+| Which Part | Player 1 | Player 2 | Player 3 | Others? |
+|---|---|---|---|---|
+| `V1` | the violist | the cellist `(vc)` | the pianist `(p)` | |
```
jsundram: **This example contradicts the rule stated two paragraphs above, and following it produces wrong stats.**

Line 53 says *"Playing V1 means Player 1 is V2"*, and line 66 says *"Anyone whose instrument isn't the one their seat implies needs an annotation"*. In this piano-quartet row the user is `V1`, so Player 1's implied seat is **V2** — but the violist sits there with no annotation.

`computePartBreakdownPerMusician` will therefore count that violist under **V2** (via `SLOT_TO_PART.V1[0]`), which is exactly the mistake the section exists to prevent. The violist needs `(va)`:

```suggestion
| `V1` | the violist `(va)` | the cellist `(vc)` | the pianist `(p)` | |
```

scripts/audit_ensembles.py
```diff
@@ -0,0 +1,183 @@
...
+            (explicit_short if stated else assumed_short).append((row, need, got))
+        if mentions_keyboard(row) and not has_keyboard_annotation(row):
+            unannotated.append(row)
+
+    print(f"Rows: {len(rows)}\n")
+
+    print(f"=== UNDER-LOGGED: title states the ensemble ({len(explicit_short)}) ===")
+    print("Someone is missing from these rows. Reconstruct only what you're sure of.\n")
+    for row, need, got in explicit_short:
+        print(f"  {(row.get('Timestamp') or '')[:10]:10s} "
```
jsundram: **`[:10]` assumes ISO dates, but the export writes `M/D/YYYY H:mm:ss`.**

`csvFormat.formatTimestamp` emits unpadded month/day/hour (`${d.getMonth()+1}/${d.getDate()}/...`), so slicing 10 characters cuts into the time for any single-digit month or day. Actual output from this script:

```
  1/1/2024 1 Mozart         Piano Quartet K478   needs 4, logged 3
  1/1/2024 2 Brahms         Sextet 1             needs 6, logged 6
```

The date column is the only thing that lets you find the row in the sheet, and it is mangled for most of the year. Same bug at line 170 (and at line 116 of `describe`'s caller block). `(row.get('Timestamp') or '').split(' ')[0]` fixes it.

src/dataProcessor.js
```diff
@@ -474,11 +517,15 @@ export function computePartBreakdownPerMusician(rows) {
...
-            [d.player1, d.player2, d.player3].forEach((name, i) => {
-                if (name && name !== '-') bump(name, slotParts[i]);
-            });
-        }
+        [d.player1, d.player2, d.player3].forEach((name, i) => {
+            if (!name || name === '-') return;
+            // An annotated slot states its own part; fall back to the seat the
+            // quartet layout implies. A row with neither contributes nothing,
+            // exactly as before.
+            const annotated = d.playerInstruments?.[i];
```
jsundram: **Slot annotations are now honored here but nowhere else, so one row gets two different answers.**

`extractUniquePlayers` (line ~808) and `checkSinglePlayerMatch` (line ~856) still classify slots purely by `SLOT_TO_PART[d.part]`.

Concrete: piano quartet logged as `part=V1`, `Player 3 = "Alice Hart (p)"`.
- Dashboard / network: `partFromInstrument('p')` → `OTHER`.
- Player dropdown: token `Alice Hart.vc`; selecting it together with the `VC` part button matches this row, i.e. the UI claims Alice played cello on the very row the charts say she played piano.

Worth routing the annotation through the same two functions so the filter and the charts agree.

scripts/audit_fillforward.py
```diff
@@ -0,0 +1,163 @@
...
+import re
+import sys
+from datetime import datetime
+from pathlib import Path
+
+SESSION_WINDOW_HOURS = 4  # mirrors SESSION_WINDOW_HOURS in src/dataProcessor.js
+CATALOG = Path(__file__).resolve().parent.parent / "static/data/all_works.json"
+# Parts that only exist in an ensemble larger than a quartet. A row naming one
+# of these in Others? is a quintet/sextet; the next row need not be.
+EXTRA_STRING_RE = re.compile(r"^(?:va|vla|vc|v)\s*[2-9]\b|^v[3-9]\b", re.I)
```
jsundram: **`v2` is a seat a string quartet *does* have, so this suppresses real findings.**

The `v` branch of `^(?:va|vla|vc|v)\s*[2-9]\b` matches `v2`, but the comment above says these are *"Parts that only exist in an ensemble larger than a quartet"* — which is true of `va2`/`vc2`/`v3+`, not of `v2`.

Concrete: anchor row is a known quartet (`Beethoven 18#1`) with all three slots filled plus `Others? = "Eve Ng (v2)"` — a fifth person rotating through. The next blank-slot row is another catalogued quartet, so `needs_the_extra_player` returns `False` and Eve's dropped appearance is never reported, even though the justification (*"a quartet has no seat for them"*) doesn't hold: she was a fifth body in the room either way.

Dropping the bare `v` alternative (`^(?:va|vla|vc)\s*[2-9]\b|^v[3-9]\b`) restricts it to parts a quartet genuinely lacks.

scripts/audit_all.sh
```diff
@@ -0,0 +1,102 @@
...
+count() { grep -E "$1" "$2" | head -1 | grep -oE '\([0-9]+\)' | tr -d '()'; }
+rule "SUMMARY"
+printf '  %-46s %s\n' \
+  "bare names with 2+ candidates (NEEDS MEMORY)" "$(count 'bare names in the sheet with 2\+ candidates' "$OUT/aliases.txt")" \
+  "aliases keyed on an ambiguous first name" "$(count 'aliases keyed on an ambiguous first name' "$OUT/aliases.txt")"
+# Deliberately not taken from the run above: on the processed export every
+# canonical name is present by construction, so the count is always 0 and
+# would read as "nothing to see". Only the raw sheet can answer it.
+printf '  %-46s %s\n' \
+  "aliases whose canonical name is absent" \
```
jsundram: **The raw-sheet pass trades one systematic error for its mirror image.**

The comment is right that the processed export always reports 0. But `collect_appearances` never applies `PLAYER_ALIASES` — it only expands `PLAYER_ABBREVIATIONS` — so on the *raw* sheet, `present` contains exactly the strings a human typed. Any alias whose canonical name exists only in the table is then always "dangling".

That is the normal shape for a spelling normalization. The current table has e.g. `"Jennifer Minnich": { cello: "Jen Minnich" }`; if the sheet only ever contains `Jennifer Minnich` and `Jen`, `Jen Minnich` appears nowhere raw and this line reports a correctly-working alias as needing repair — the same "you'll go delete a live alias" failure `report_ambiguity`'s own guard was added to prevent.

A canonical name that is the target of some alias should probably count as present.

PR #19, opened 2026-08-28
Attribute ambiguous bare names from the row, not from memory

jsundram:
`report_ambiguity` listed bare first names and the full names each could match: 11 names, 32 entries, all marked NEEDS MEMORY. The name alone genuinely is a memory problem. The row is not — two people who share a first name rarely share a stand, so the other players identify which one was there.

`attribute_bare_entries` scores each candidate by how many of the row's other players fall in that candidate's teammate circle, and reports only what means something:

- **the room contradicts the alias** — that entry is credited to the wrong person today
- **the room resolves an entry no alias covers** — the app counts that bare form as a separate person; mechanical to fix
- **nobody has decided** — no alias, no evidence: the only kind that decays

Entries that agree with the table, and entries with an alias standing that this run cannot second-guess, are counted rather than listed.

Review found the first version was reading its evidence from data the alias under test had already rewritten, so a wrong alias confirmed itself. It reads `archive/data-raw.csv` now, warns loudly if handed a canonicalized export, and `audit_all.sh` takes these counts from the raw pass.

Guards, each added because the unguarded version produced a wrong finding on the real log:

- circles and `written` counts come only from names a human typed out — not alias-supplied, not fill-forwarded
- `MIN_WRITTEN_IN_FULL` on the winner **and** on any alias it contradicts **and** at least one rival: the people most often logged bare are the ones whose full name is rarest, so an unnamed rival's silence means "never seen", not "not them". Rivals that cannot be ruled out are printed, not discarded.
- candidates are class-keyed like `PLAYER_ALIASES`, and a one-letter last token is an abbreviation rather than a rival person
- the subject's own cell is excluded positionally, since comparing by name fails once fill-forward has expanded it
- `candidate_index` is the single definition of who a bare name could be, shared with the per-name list

On the real log: **2 contradicted, 1 unaliased, 0 undecided.** Two of those were fixed in the sheet during review; the remaining two rest on a single shared teammate and name a rival that could not be ruled out, so they are "check this", not "this is wrong".

## Also

- `fill_forward` runs the real `fillForward` over the raw sheet via node with empty tables, returning every row twice: the filled row is the evidence, the written row is the only one with a cell anyone could edit
- `audit_all.sh` no longer discards the raw pass's stderr or swallows its exit status; a degraded fill-forward would otherwise inflate NEEDS MEMORY silently

320 tests, lint and typecheck clean. No real names in the diff.

## Tests

93 pytest cases in `test/test_audits.py`, gated in CI — the audit logic had none, and every finding in this PR's four review rounds was in it. pytest rather than a hand-rolled runner because `audit_aliases` resolves `PLAYER_ALIASES` at *import* time from the gitignored `src/aliases.js`: without an injected table a test reads real names locally and the empty stub in CI. An autouse `monkeypatch` fixture makes that impossible to forget. Nothing enters `package.json` — uv fetches a pinned pytest per run, with the same uv-or-python3 fallback `audit_all.sh` uses.

The suite ends with **mirror tests** running each hand-ported Python helper against the JavaScript it copies. That caught real drift: `class_of` recognised only `vc`, while the app's `classOf` also takes `cello`, `violoncello`, `vlc`, `c` — so a cellist written `(cello)` was classed as an upper player by the audit and a cellist by the app, sending it to the wrong half of a class-keyed alias.

Every guard was then mutation-tested — remove it, confirm a test goes red. That found two tests worth less than they looked: one named for the subject's-own-seat exclusion was really exercising the sheet-resolved skip, and the contradicted-alias gate was invisible because with two candidates the rival gate subsumes it. Both are now isolated.

## Split out

The `fillForward` semantics change this branch had grown is now #20, stacked on this one. It edits the data layer and moves user-visible stats; this PR only reads the sheet.


scripts/audit_aliases.py
```diff
@@ -226,7 +240,71 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+                 entry here that is ever a bug, and the reason to run this.
+      unsettled  no candidate's circle matches — a one-off group, a reading
+                 party. These genuinely need memory, and they are the only
+                 ones that decay.
+      settled    count of rows the teammates settle in agreement with the
+                 alias. Not listed: nothing to do, and listing them buries the
+                 two lists above.
+    """
+    # Circle = everyone this person has ever played with, either class.
+    circles: dict[str, set[str]] = defaultdict(set)
```
jsundram, line 404: **The evidence for attribution comes from data the alias under test has already rewritten, so a wrong alias confirms itself.**

`circles` is built from `appearances`, which is built from the input CSV — by default `archive/data.csv`, the *processed* export, whose Player 1/2/3 slots have already been through `normalizePlayerNames`. If `PLAYER_ALIASES['Alice'].upper == 'Alice Hart'` is wrong for 200 slot rows, those 200 rows contributed their teammates to `circles['Alice Hart']`. When the remaining bare `Alice` rows are then scored, Alice Hart's (contaminated) circle wins and the row is filed as `settled` — "agrees with the table". The one case the section exists to catch is exactly the case the input is biased against surfacing. The same contamination underlies the "jaccard 0.00 for every ambiguous pair" measurement quoted in the docstring and CLAUDE.md: post-canonicalization the circles are disjoint partly *because* the alias split them.

Attribution wants the raw sheet (`archive/data-raw.csv`), where no alias has been applied yet, or at minimum a loud caveat that a `conflicts` count of 0 on the processed export is not evidence.

scripts/audit_aliases.py
```diff
@@ -226,7 +240,71 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+        people = row_people(row, slot_classes)
+        names = [n for n, _ in people]
+        for name, cls in people:
+            candidates = full_by_first.get(base_token(name), set())
+            if len(name.split()) > 1 or len(candidates) < 2:
+                continue
+            mates = [m for m in names if m != name]
+            scored = sorted(((len([m for m in mates if m in circles[c]]), c)
+                             for c in candidates), reverse=True)
+            top, runner = scored[0], (scored[1] if len(scored) > 1 else (0, ""))
```
jsundram: **On the default input, `conflicts` can only ever fire for `Others?` entries — a bare name in a player slot is unreachable.**

`alias = EXISTING_ALIASES.get(name, {}).get(cls)` is looked up on a name read out of the CSV. In `archive/data.csv` the three player slots are canonicalized (`normalizePlayerNames`) while `Others?` is exported verbatim (`rowToFields` emits `d.others`, not `d.othersList`). So any bare slot name that has an alias has already been replaced by its canonical full name and no longer matches the `len(name.split()) > 1` guard at line 283; `alias` is `None` for every slot-derived person, and the `elif alias and ...` branch is dead for them.

Concretely: the mis-attributed rows the docstring calls "the reason to run this" are invisible for the seats where most people are logged. `scripts/audit_all.sh:87` compounds it by pulling the count from the processed run, so the summary line will read 0 regardless of how wrong the table is.

scripts/audit_aliases.py
```diff
@@ -226,7 +240,71 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            alias = EXISTING_ALIASES.get(name, {}).get(cls)
+            if not top[0] or top[0] == runner[0]:
+                unsettled.append((row, name, cls, alias, sorted(candidates)))
+            elif alias and alias != top[1]:
+                conflicts.append((row, name, cls, alias, top[1],
+                                  [m for m in mates if m in circles[top[1]]]))
+            else:
+                settled += 1
```
jsundram, line 551: **`settled` lumps together two different situations, and the summary line describes both as "nothing to do".**

This `else` is reached both when the room's winner matches an existing alias *and* when there is no alias at all (`alias is None`, room picked a winner). The second case is not settled in any actionable sense: section 1 above reports exactly those names as `NO alias — counted as its own person`, and the hazard-1 text says such rows are unfixable by an alias and need the SHEET edited. Yet line 411 prints them as "settled by the room and agree with the table — nothing to do", and `audit_all.sh` only surfaces `unsettled` as work.

Scenario: a bare `Alice` with no alias appears in 40 rows, the room resolves each one to `Alice Hart`. The audit says there is nothing to do, while the app still counts a phantom person "Alice" separately from "Alice Hart" in every people/collaborator stat, and the audit has just computed the answer needed to fix the sheet but discards it. Suggest splitting a third bucket (room-resolved, no alias) and listing it — it is the largest source of real, non-decaying work this function can find.

scripts/audit_aliases.py
```diff
@@ -226,7 +240,71 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+        if len(name.split()) > 1:
+            full_by_first[base_token(name)].add(name)
+
+    conflicts, unsettled, settled = [], [], 0
+    slot_classes = slot_classes_for(rows)
+    for row in rows:
+        people = row_people(row, slot_classes)
+        names = [n for n, _ in people]
+        for name, cls in people:
+            candidates = full_by_first.get(base_token(name), set())
```
jsundram: **Candidate selection ignores instrument class, which is the one axis `PLAYER_ALIASES` is keyed on.**

`full_by_first` (line 271) is built by discarding `_cls`, and `circles` (267) merges both classes, so the candidates for a bare name include full names who have never played that class. Given the table's motivating example from CLAUDE.md — `{"Jo": {upper: "Jo Alpha", cello: "Jo Beta"}}` — a bare `Jo` in the cello slot draws candidates `{Jo Alpha, Jo Beta}`, and if Jo Alpha's (upper, much larger) circle contains more of the row's mates, `top[1] == 'Jo Alpha'` while `alias == 'Jo Beta'` and the row is reported as a conflict. The correct class-keyed alias is flagged as crediting the wrong person.

Restricting candidates to full names actually seen in `cls` (i.e. keep the class in `full_by_first`) removes this class of false positive.

scripts/audit_aliases.py
```diff
@@ -226,7 +240,71 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+    for (name, _cls), apps in appearances.items():
+        circles[name].update(teammate_counter(apps))
+
+    full_by_first: dict[str, set[str]] = defaultdict(set)
+    for (name, _cls) in appearances:
+        if len(name.split()) > 1:
+            full_by_first[base_token(name)].add(name)
+
+    conflicts, unsettled, settled = [], [], 0
+    slot_classes = slot_classes_for(rows)
```
jsundram: `slot_classes_for(rows)` is invoked a second time here — `collect_appearances` already ran it on the same `rows` earlier in `main`. Each call spawns a `node -e` subprocess over every annotated slot value in the file, so the whole-file node round-trip is paid twice, and on a node failure the "could not class slot annotations" note is printed twice for one run. Passing the map in (or memoizing `slot_classes_for`) would avoid it.

scripts/audit_aliases.py
```diff
@@ -226,7 +240,71 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            candidates = full_by_first.get(base_token(name), set())
+            if len(name.split()) > 1 or len(candidates) < 2:
+                continue
+            mates = [m for m in names if m != name]
+            scored = sorted(((len([m for m in mates if m in circles[c]]), c)
+                             for c in candidates), reverse=True)
+            top, runner = scored[0], (scored[1] if len(scored) > 1 else (0, ""))
+            alias = EXISTING_ALIASES.get(name, {}).get(cls)
+            if not top[0] or top[0] == runner[0]:
+                unsettled.append((row, name, cls, alias, sorted(candidates)))
```
jsundram: Both lists are appended once per ambiguous *person occurrence*, not once per row, but every label calls them "bare rows" (headers at 391/402, and the `audit_all.sh` summary). A row containing two different ambiguous bare names — not rare in the `Others?` column of a camp session — is counted twice, so the NEEDS MEMORY number the summary presents overstates the number of rows to look at. Either dedupe by row id or rename to "bare entries".

scripts/audit_aliases.py
```diff
@@ -226,7 +253,108 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+                             for c in candidates), reverse=True)
+            top, runner = scored[0], scored[1]
+            alias = EXISTING_ALIASES.get(name, {}).get(cls)
+            why = [m for m in mates if m in circles[top[1]]]
+            # A circle is only evidence if we have one. Someone almost always
+            # logged bare has a thin explicit circle, so their FAILURE to match
+            # says nothing — and the rival who happens to share one hub mate
+            # would win by default. Both the winner and, when contradicting it,
+            # the alias's target must be people the sheet actually names.
+            confident = (written[top[1]] >= MIN_WRITTEN_IN_FULL
```
jsundram, line 536: **The confidence gate never checks the runner-up's attestation, so a thinly-written rival is silently ruled out.**

The comment right above says the hazard is that "someone almost always logged bare has a thin explicit circle, so their FAILURE to match says nothing — and the rival who happens to share one hub mate would win by default." But `confident` only tests `written[top[1]]` (the winner) and, when it disagrees, `written[alias]`. Nothing tests the losing candidate.

Reproduced on a fixture: `Alice Hart` written out 6x with `Bob Zeta`/`Carl Yun`; `Alice Bek` written out exactly once (circle `{Dora Xu, Ed Wu}`); one bare `Alice` row whose mates are `Carl Yun, Fay Vo`. Scores are Hart 1, Bek 0, `written[Hart]=6` so `confident` is true, and the run prints

```
'Alice' [upper] -> 'Alice Hart'  (played with Carl Yun)
```

in the *"the room resolves but no alias covers"* bucket, whose copy tells you this is "mechanical: write the full name into the cell". Bek's non-match carries no information at all here — she has been named once — yet the output invites an irreversible edit to the sheet crediting the wrong person. Since `unaliased` is described as the largest actionable bucket, this is the finding most likely to cause a bad edit.

Gating on the runner-up too (e.g. `written[runner[1]] >= MIN_WRITTEN_IN_FULL`, or requiring a margin proportional to the rivals' attestation) would push these into `unsettled`, which is where "we genuinely cannot tell" belongs.

scripts/audit_aliases.py
```diff
@@ -172,30 +178,51 @@ def load_rows(path: Path) -> list[dict]:
...
     }
-    slot_classes = slot_annotation_classes(annotated)
+    return slot_annotation_classes(annotated)
+
+
+def row_people(row: dict, slot_classes: dict[str, str]) -> list[tuple[str, str]]:
+    """(name, class) for everyone in one row.
+
+    The single reader of a row's cast, shared by the teammate index below and
+    by attribute_bare_rows, so the two cannot disagree about who was there.
```
jsundram: Stale function name in the docstring: the function is `attribute_bare_entries` (renamed in this commit), not `attribute_bare_rows`. Grepping for the name from this comment finds nothing.

scripts/audit_aliases.py
```diff
@@ -226,7 +253,108 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            # would win by default. Both the winner and, when contradicting it,
+            # the alias's target must be people the sheet actually names.
+            confident = (written[top[1]] >= MIN_WRITTEN_IN_FULL
+                         and (not alias or alias == top[1]
+                              or written.get(alias, 0) >= MIN_WRITTEN_IN_FULL))
+            if not top[0] or top[0] == runner[0] or not confident:
+                if alias:
+                    unverified += 1
+                else:
+                    unsettled.append((row, name, cls, alias, sorted(candidates)))
```
jsundram: `alias` is always `None` on this branch — the `if alias:` immediately above routes every aliased entry into `unverified`. So the 5th tuple element is dead, and the printer at the bottom (`f"  alias -> {alias or 'none'}"`) unconditionally renders the literal string `alias -> none` for every unsettled entry. Either drop the field and the printed suffix, or the printed line is claiming to report something it can never vary.

scripts/audit_aliases.py
```diff
@@ -226,7 +253,108 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+                           slot_classes: dict[str, str],
+                           ) -> tuple[list, list, list, int, int]:
+    """Decide, per entry, which of several same-first-name people a bare name is.
+
+    "Which Alice was this" reads as a memory problem, and the name alone makes
+    it one. The row is not alone though: two people who share a first name
+    rarely share a stand, so the OTHER people in the row point at the one who
+    was there, long after anyone could recall the evening.
+
+    MUST be measured on the RAW sheet (scripts/fetch_raw.sh -> data-raw.csv).
```
jsundram, line 442: The mandatory raw-sheet input costs evidence that `fillForward` would supply, and the loss lands exactly in the bucket labelled NEEDS MEMORY.

On `archive/data-raw.csv` today, 1316 of 3389 rows have all three player slots blank (continuation rows), and 35 of those still carry an `Others?` value. For those 35, `row_people()` returns only the `Others?` entries, so an ambiguous bare name there gets a `mates` list of at most the other `Others?` people — usually empty. `top[0]` is then 0, and the entry is filed as `unsettled` ("nobody has decided ... only you can close ... they decay") or `unverified`, even though the anchor row directly above names the whole quartet.

The reason for demanding raw input is `normalizePlayerNames`, not `fillForward`; running `prepareRows` + `fillForward` on the raw sheet *without* the name normalization would keep the evidence honest and recover these rows.

scripts/audit_aliases.py
```diff
@@ -226,7 +253,108 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            full_by_first[(base_token(name), cls)].add(name)
+
+    conflicts, unaliased, unsettled, unverified, settled = [], [], [], 0, 0
+    for row in rows:
+        people = row_people(row, slot_classes)
+        names = [n for n, _ in people]
+        for name, cls in people:
+            if len(name.split()) > 1:
+                continue
+            candidates = full_by_first.get((base_token(name), cls), set())
```
jsundram: This candidate set is class-keyed and excludes abbreviated forms (`Peter O`), but section 1's `full_by_first` a few lines below (`report_ambiguity`, built with a bare `base_token(name)` key and no abbreviation guard) is neither. The two sections therefore disagree about the same name in the same output: on `archive/data-raw.csv` section 1 lists 17 bare names "with 2+ candidates", while 1b silently `continue`s on any of them whose rivals differ only by class or are just abbreviations, so those names appear in the "Fix these in the SHEET" list and then never again. Sharing one candidate-builder (as `row_people` now does for the cast) would keep the two counts explicable to a reader.

scripts/audit_aliases.py
```diff
@@ -226,7 +297,132 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+        # Evidence from the filled row, subjects from the row as written: a
+        # continuation row's blank slot is not a cell anyone can go and fix.
+        names = [n for n, _ in row_people(filled, slot_classes)]
+        for name, cls in row_people(row, slot_classes):
+            if len(name.split()) > 1:
+                continue
+            candidates = full_by_first.get((base_token(name), cls), set())
+            if len(candidates) < 2:
+                continue
+            mates = [m for m in names if m != name]
```
jsundram: **The entry being attributed is counted as its own stand-mate, and can flip the verdict.**

`names` comes from the *filled* row, `name` from the row *as written*. When `fillForward` resolved this very slot — a same-session bare `Peter` after `Peter Chan` matches `refersToPrevEntry`’s prefix-at-word-boundary rule — the filled slot holds `Peter Chan` while `name` is still `Peter`, so `m != name` does not remove it. The subject then "played with" itself, and the point goes to whichever *rival* has `Peter Chan` in their circle, never to Peter Chan.

Reproduced on a synthetic raw sheet (Peter Chan ×5 w/ Bob Ray; Peter Krauss ×5, once alongside Peter Chan; then a session `Peter Chan, Bob Ray` followed 1h later by `Peter, Bob Ray`), with `PLAYER_ALIASES = {"Peter": {"upper": "Peter Chan"}}`:

```
CONFLICTS:
  3/1/2024 11:00:00 Peter  alias Peter Chan -> room says Peter Krauss
                    because [Peter Chan, Bob Ray]
```

The output cites *Peter Chan* as the reason it was not Peter Chan, and asks you to edit the sheet cell to the wrong name — on a row where the sheet’s own repetition had already answered it. Two things to fix: drop the subject’s own seat from `mates` (needs the slot index / a positional exclusion rather than a name compare), and treat the fill-forwarded value of that slot as evidence rather than as a teammate.

scripts/audit_aliases.py
```diff
@@ -226,7 +297,132 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+    """
+    by_first: dict[tuple[str, str], set[str]] = defaultdict(set)
+    circles: dict[str, set[str]] = defaultdict(set)
+    written: Counter = Counter()
+    for (name, cls), apps in appearances.items():
+        tokens = name.split()
+        if len(tokens) > 1 and len(tokens[-1].rstrip(".")) > 1:
+            by_first[(base_token(name), cls)].add(name)
+            circles[name].update(teammate_counter(apps))
+            written[name] += len(apps)
```
jsundram, line 409: **`written` counts fill-forwarded rows, so `MIN_WRITTEN_IN_FULL` is met by a name typed once.**

`appearances` is built in `main()` from the *filled* rows (`rows = [filled for _written, filled in pairs]`), so `len(apps)` counts every continuation row `fillForward` populated, not the times a human wrote the name out. The constant’s own comment says the gate exists because "the people most often logged bare are exactly the ones whose full name is rarest" — this is precisely that population, and session repetition inflates them straight through it.

Synthetic raw sheet: `Alice Hart` typed on row 1 of a 5-piece session with 4 blank continuation rows, `Alice Bek` likewise in a second session, then one bare `Alice`:

```
written: {"Alice Hart": 5, "Bob Ray": 6, "Cara Vine": 6, "Alice Bek": 5, ...}
```

Each Alice was typed **once**, yet both clear `written >= 5`, `confident` is True, and the run emits a firm attribution (`unaliased: Alice -> Alice Hart`) from a single typed occurrence each. Same root affects `circles` on the line above, which the docstring claims are built from "names a human typed in full".

Counting distinct *written* occurrences (i.e. building `written`/`circles` from the as-written rows in `pairs`, keeping the filled rows only for the per-entry evidence lookup) would restore the intended meaning.

scripts/audit_aliases.py
```diff
@@ -374,8 +625,12 @@ def main() -> None:
         print(f"CSV not found: {csv_path}", file=sys.stderr)
         sys.exit(1)
 
-    rows = load_rows(csv_path)
-    appearances = collect_appearances(rows)
+    # (row as written, row after fill-forward) — see fill_forward.
+    pairs = fill_forward(load_rows(csv_path))
+    rows = [filled for _written, filled in pairs]
```
jsundram: **Feeding the filled rows to `collect_appearances` silently changes every pre-existing report — and makes hazard 1 miss bare cells that are in the sheet.**

Before this PR `appearances` came from the rows as written. Now it comes from post-`fillForward` rows, so `bare_count`, `present`, `sheet_surnames`, the variant/teammate counts and the canonicalized-export detector all read values `fillForward` synthesised rather than what a human typed. The attribution feature only needs the filled rows for the *evidence* side; the index does not.

The concrete regression: `fillForward` expands a same-session bare name that prefixes the previous entry, so those cells disappear from `bare_count`. Same synthetic raw sheet as my other comment (`Peter Chan, Bob Ray` at 10:00, `Peter, Bob Ray` at 11:00):

```
as written (old behaviour) bare: {("Peter", "upper"): 1}
filled     (new behaviour) bare: {}
```

Section 1 ("bare names in the sheet with 2+ candidates … Fix these in the SHEET") therefore omits a cell that literally reads `Peter`, while section 1b — which reads the as-written row — reports the very same cell. That is exactly the disagreement `candidate_index`’s docstring says it exists to prevent, and it lands on `archive/data-raw.csv`, the input `audit_all.sh` now pulls the headline numbers from.

Suggest building `appearances`/`candidate_index` from `[w for w, _f in pairs]` and passing the filled rows only where the evidence is read.

scripts/audit_all.sh
```diff
@@ -90,6 +90,14 @@ n_of() { printf '%s' "$RAWALIAS" | grep -E "$1" | head -1 | grep -oE '\([0-9]+\)
 printf '  %-46s %s\n' \
   "aliases that are a surname's only record" "$(n_of 'ONLY record of a surname') (back up src/aliases.js)" \
   "aliases pointing at an unrelated name" "$(n_of 'absent and unrelated')"
+# Also from the raw pass, and for a sharper reason: attribution reads a bare
+# name's teammates as evidence, and on the processed export those teammates
+# were themselves supplied by the alias under test. A wrong alias would vote
+# to confirm itself, so a 0 there is not evidence of anything.
+printf '  %-46s %s\n' \
+  "bare entries nobody has decided (NEEDS MEMORY)" "$(n_of 'bare entries nobody has decided')" \
```
jsundram, line 133: **These three counts can be silently wrong, because the raw pass’s stderr is discarded.**

`RAWALIAS` is captured with `2>/dev/null || true` (line 88). The new `fill_forward()` degrades on any node failure — a short CSV row makes `DictReader` emit `None`s, which survive `processRow`’s `d[c] === undefined` guard and then blow up on `.trim()`; node exits 1; Python catches it and returns `[(row, row)]` after printing `Note: could not fill-forward via node …` to **stderr**. That note is thrown away here, and with no fill-forward every continuation row loses its cast, so `bare entries nobody has decided (NEEDS MEMORY)` inflates with rows whose answer is sitting in the row above — the opposite of what the flag means, presented with no signal that anything went wrong.

The same `|| true` means a raw pass that fails outright prints these three labels with empty values rather than an error.

Worth either letting the raw pass’s stderr through, or having `fill_forward` mark the degradation on stdout so the summary can say so.

scripts/audit_all.sh
```diff
@@ -80,16 +80,33 @@ rule "DROPPED BY FILL-FORWARD   scripts/audit_fillforward.py"
...
   "aliases keyed on an ambiguous first name" "$(count 'aliases keyed on an ambiguous first name' "$OUT/aliases.txt")"
 # Taken from the raw sheet, not the run above: on a canonicalized export
 # every live alias's canonical name is present by construction, so both
 # counts collapse to near-nothing there and would read as "nothing to see".
-RAWALIAS=$("${PY[@]}" scripts/audit_aliases.py archive/data-raw.csv 2>/dev/null || true)
+# stderr is NOT discarded and failure is NOT swallowed: this pass runs
+# fill_forward, which degrades to un-filled rows if node fails, and a degraded
+# run inflates the NEEDS MEMORY count with rows whose answer sits in the row
+# above. Silently printing that as a finding is worse than printing nothing.
+if ! RAWALIAS=$("${PY[@]}" scripts/audit_aliases.py archive/data-raw.csv); then
```
jsundram: **The raw pass's report body is thrown away, so the new findings can only be counted, never read.**

`RAWALIAS` is captured into a shell variable and only ever grepped for section headers — nothing `tee`s it. The three new summary rows ("contradicts the room", "resolved but not aliased", "nobody has decided") therefore report counts whose supporting lines (date/composer/work, the cell, the winner, the rivals that could not be ruled out) exist only inside that discarded string.

Scenario: `npm run audit` prints `bare entries whose alias contradicts the room   2`. The user now wants to fix those two sheet cells — but the displayed `PLAYER NAMES` section came from the *processed* pass, whose own attribution block is printed under the "measured on the CANONICALIZED export … a 0 here is not evidence" banner. There is no path from the summary to the rows short of re-running `python scripts/audit_aliases.py archive/data-raw.csv` by hand, which the summary doesn't say to do.

Suggest `tee`ing the raw pass to `$OUT/aliases-raw.txt` and either printing its 1b sections or naming the file in the summary.

scripts/audit_all.sh
```diff
@@ -80,16 +80,33 @@ rule "DROPPED BY FILL-FORWARD   scripts/audit_fillforward.py"
...
+    echo "  !! the raw-sheet pass FAILED — the counts below are missing, not zero." >&2
+fi
+if printf '%s' "$RAWALIAS" | grep -q 'could not fill-forward'; then
+    echo "  !! fill-forward degraded on the raw pass — NEEDS MEMORY below is inflated." >&2
```
jsundram: **These two warnings go to stderr while everything they qualify goes to stdout.**

`fill_forward` deliberately prints its degradation notice on *stdout* ("this lands inside the report a caller captures, so a degraded run cannot be mistaken for a clean one") — and then `audit_all.sh` re-emits it on stderr, undoing that reasoning at the level the user actually reads.

Scenario: the documented redirect workflow (`alias-output.txt` is gitignored for exactly this in CLAUDE.md) — `npm run audit > report.txt`. Node is missing or `processRow` throws on one row; the raw pass degrades. `report.txt` contains the summary with an inflated NEEDS MEMORY count and no trace of either warning, because both went to the terminal. Same for the harder failure: if the raw pass exits non-zero, the summary rows render empty and the file gives no reason why.

Both echoes should go to stdout, next to the numbers they are about.

scripts/audit_aliases.py
```diff
@@ -172,32 +187,112 @@ def load_rows(path: Path) -> list[dict]:
...
+        "let s='';process.stdin.on('data',d=>s+=d).on('end',()=>Promise.all(["
+        "import('./src/dataProcessor.js'),import('./src/csvFormat.js')])"
+        ".then(([dp,cf])=>{const {rows}=dp.prepareRows(JSON.parse(s).map(dp.processRow));"
+        "const before=cf.serializeRows(rows);dp.fillForward(rows,{});"
+        "process.stdout.write(JSON.stringify({before,after:cf.serializeRows(rows)}));})"
+        ".catch(e=>{console.error(e.message);process.exit(1);}))"
+    )
+    try:
+        result = subprocess.run(
+            ["node", "-e", js], input=json.dumps(rows),
```
jsundram: **One malformed row degrades fill-forward for the whole file.**

`csv.DictReader` fills missing trailing fields with `None` (`restval`), and `processRow`'s guard tests `d[c] === undefined`, so `null` slips past it and the first `.trim()` throws. Node exits 1, the `except` fires, and *every* row in the file loses its filled view.

Reproduced on a two-row CSV whose second line is short by three fields:

```
  !! could not fill-forward via node (Row is missing expected column(s): Others?) — continuation rows
     lost their cast, so the counts below understate what is
     answerable and overstate what needs memory.
```

The blast radius is the issue, not the failure: on the real log roughly a third of rows are continuations, so one truncated line converts most of the sheet's answerable entries into `unsettled` — the bucket `audit_all.sh` labels NEEDS MEMORY. Cheap fix at the boundary:

```suggestion
        result = subprocess.run(
            ["node", "-e", js],
            input=json.dumps([{k: (v or "") for k, v in r.items() if k} for r in rows]),
```

scripts/audit_aliases.py
```diff
@@ -172,32 +187,112 @@ def load_rows(path: Path) -> list[dict]:
...
+            # An instrument annotation states the class; the column only
+            # implies it (SLOT_CLASS). Same precedence as the app.
+            cls = slot_classes.get(raw) or SLOT_CLASS[i]
+            people.append((expand_abbrev(strip_parens(raw)), cls, f"p{i + 1}"))
+    # The canonical header is "Others?" (see src/csvFormat.js), but exports
+    # written before the header fix used "Others" — accept both.
+    for i, (name, instr) in enumerate(
+            parse_others(row.get("Others?") or row.get("Others") or "")):
+        cls = class_of(instr)
+        if name and cls:
```
jsundram: **An `Others?` entry with no `(instrument)` is dropped here, and that now costs attribution twice.**

`parse_others` explicitly supports a bare `Name` (its docstring lists `Name`, `Name (instrument)`, `Name (instrument, comment)`), but `class_of(None)` is `None`, so those entries never make it into `people`. Two consequences for `attribute_bare_entries`, both verified on synthetic rows:

1. **Invisible as a subject.** A row with `Others?: "Alice"` and two candidate Alices yields `([], [], [], 0, 0)` — the entry appears in no bucket at all, not even `unsettled`. The app *does* count that bare form as its own person, so the summary's NEEDS MEMORY figure silently understates.
2. **Missing as evidence.** With `Others?: "Beryl"` (unannotated) as the only stand-mate who could identify the subject, the entry lands in `unsettled` — a false NEEDS MEMORY, which is the exact failure this PR exists to remove. Adding `(v2)` to Beryl moves the same row to `unaliased` with the right winner.

A name with no instrument still tells you who was in the room. Consider keeping unannotated others in the cast (falling back to `SLOT_CLASS`-free "unknown" for the class, or only excluding them from *candidate/class* keying rather than from `people` entirely).

scripts/audit_aliases.py
```diff
@@ -73,11 +74,25 @@ def load_alias_tables() -> tuple[dict[str, dict[str, str]], dict[str, str]]:
...
+# in test/test_audits.py, which is what caught this reading "(cello)" as an
+# upper part while the app read it as a cellist.
+CELLO_INSTRUMENT = re.compile(r"^(?:vc|vlc|cello|violoncello|c)(?![a-z])", re.I)
+
 
 def class_of(instrument: str | None) -> str | None:
     if not instrument:
         return None
-    return "cello" if instrument.lower().strip().startswith("vc") else "upper"
+    s = instrument.lower().strip().replace("asst ", "").replace("ast ", "")
```
jsundram: **The assistant-prefix strip diverges from the app it mirrors, and the mirror test's sample list doesn't cover the gap.**

JS is `s.toLowerCase().trim().replace(/^as?st\s+/, '')` — anchored, any whitespace run. This is `.replace("asst ", "").replace("ast ", "")` — unanchored, one literal space. Verified against the real `classOf`:

| input | `aa.class_of` | `m.classOf` |
| --- | --- | --- |
| `"asst  vc"` (two spaces) | `upper` | `cello` |
| `"ast\tvc"` | `upper` | `cello` |

`test_class_of_matches_the_app` only samples `"asst v2"`, where both answers are `upper` either way, so the drift the mirror block was added to catch survives it. Scenario: `Bob (asst  vc)` in an `Others?` cell is keyed `upper` by the audit and `cello` by the app — so the candidate set, the alias proposal and `already_aliased` all land in the wrong half of a class-keyed table, which is the same bug the `(cello)` fix in this PR repaired.

```suggestion
    s = re.sub(r"^as?st\s+", "", instrument.lower().strip())
```
(and add `"asst  vc"` to the mirror samples)

scripts/audit_aliases.py
```diff
@@ -226,7 +321,145 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            # expands a bare name that abbreviates the previous entry in the
+            # session, and the app runs fillForward BEFORE normalizePlayerNames,
+            # so no alias ever sees this cell. Nothing to report.
+            resolved = by_seat.get(seat, name)
+            if resolved != name and base_token(resolved) == base_token(name):
+                settled += 1
+                continue
+            mates = [n for n, _c, s2 in cast if s2 != seat]
+            scored = sorted(((len([m for m in mates if m in circles[c]]), c)
+                             for c in candidates), reverse=True)
```
jsundram: **`mates` is a list, so one person occupying two cells votes twice.**

If someone is written both in a player slot and in `Others?` — which happens on the rows that push people out of the quartet layout, exactly where ambiguous bare names cluster — they contribute 2 to that candidate's score and appear twice in `why`. Verified: a row with `Player 2: "Beryl"` and `Others?: "Beryl (v2); Chantal (vc)"` produces `why == ['Beryl', 'Beryl']` and prints "played with Beryl, Beryl".

Here it only made the output look odd, but the score is what decides the verdict. Two distinct mates from candidate A's circle vs one duplicated mate from candidate B's ties at 2 → `top[0] == runner[0]` → the entry is dumped into `unsettled`/NEEDS MEMORY, when a de-duplicated count would have settled it for A. `set(mates)` (or dedupe on name when building `cast`) makes the score mean "how many of this candidate's circle were here", which is what the docstring claims it measures.

scripts/audit_aliases.py
```diff
@@ -226,7 +321,145 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            candidates = full_by_first.get((base_token(name), cls), set())
+            if len(candidates) < 2:
+                continue
+            # The sheet may already have answered this itself: fill-forward
+            # expands a bare name that abbreviates the previous entry in the
+            # session, and the app runs fillForward BEFORE normalizePlayerNames,
+            # so no alias ever sees this cell. Nothing to report.
+            resolved = by_seat.get(seat, name)
+            if resolved != name and base_token(resolved) == base_token(name):
+                settled += 1
```
jsundram: **This increments the same counter that gets reported as "agree with the table", though no table entry was consulted.**

This branch is the fill-forward-resolved skip: the sheet answered the cell itself, and `alias` hasn't even been looked up yet. It shares `settled` with line 448, which really is "the room and the alias agree", and the closing line prints the sum as:

> `({settled} more bare entries agree with the table, and {unverified} have an alias standing …)`

On the raw sheet — where about a third of rows are continuation rows and the prefix rule expands `Alice` → `Alice Hart` in-session — this branch is likely the dominant contributor, so the report tells the reader the alias table is doing work that fill-forward did. Worth a separate counter (e.g. `resolved_by_sheet`) and its own clause, since the two say different things about whether `src/aliases.js` is earning its keep.

scripts/run_audit_tests.sh
```diff
@@ -0,0 +1,30 @@
...
+node scripts/ensure_aliases.mjs
+
+if command -v uv >/dev/null 2>&1; then
+    exec uv run --with "pytest==$PYTEST_VERSION" --python 3.12 \
+        pytest test/test_audits.py "$@"
+fi
+
+echo "note: uv not found, falling back to python3 -m pytest" >&2
+python3 -c "import pytest" 2>/dev/null \
+    || python3 -m pip install --quiet --disable-pip-version-check "pytest==$PYTEST_VERSION"
```
jsundram: **This fallback is the path CI actually takes, and it installs into the runner's system Python.**

`ubuntu-latest` does not ship `uv`, and `test.yml`'s new "Run audit tests" step adds no Python setup — so in CI the script always falls through to here. On Ubuntu 24.04 the system interpreter is PEP-668 marked, where `python3 -m pip install` exits with `error: externally-managed-environment`; with `set -euo pipefail` that fails a required check on every PR. Please verify on a real run, and if it passes today it is still a system-Python mutation one image update away from breaking.

It is also the only unpinned dependency in a repo that pins every action to a commit SHA, esbuild to an exact version, and the pandoc .deb to a sha256 — `pytest==9.1.1` is fetched from PyPI at CI time with no hash.

Pinned `astral-sh/setup-uv` in `test.yml` (keeping this script's fallback for dev machines), or `python3 -m venv "$TMPDIR/…"` here, removes both problems.

scripts/audit_aliases.py
```diff
@@ -226,7 +339,165 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            if len(name.split()) > 1:
+                continue
+            if cls is None:
+                # An unannotated Others? entry: the app cannot alias it either
+                # (canonicalize with a null class is a no-op), so every class's
+                # namesakes are in play and only editing the cell can fix it.
+                candidates = {c for (token, _cls), names in full_by_first.items()
+                              if token == base_token(name) for c in names}
+            else:
+                candidates = full_by_first.get((base_token(name), cls), set())
```
jsundram: **The alias's own target is never scored, so a nickname alias reads as a contradiction.**

`candidates` is only the set of full names sharing the bare name's *first token*, but the verdict below compares `top[1]` against `alias = EXISTING_ALIASES.get(name, {}).get(cls)` — which may be a canonical name with a *different* first token. CLAUDE.md says the table exists for "anyone logged by first name only, nicknames included", so `"Nick" -> "Nicholas Hart"` is exactly the shape it is meant to hold. That person then never competes for the row, and a row the alias covers **correctly** lands in the highest-severity `conflicts` bucket, whose copy tells the reader to go edit the sheet cell.

Reproduced against this branch — aliases `{"Nick": {"upper": "Nicholas Hart"}}`; sheet has `Nick Adams` (mates Dexter/Carol, 5x), `Nick Bailey` (Chantal/Fernand, 5x), `Nicholas Hart` (Dexter/Gaston, 5x), plus one bare row `Nick / Dexter / Gaston`:

```
conflicts: [('Nick', 'Nicholas Hart', 'Nick Adams', ['Dexter'])]
```

"alias says 'Nicholas Hart', the room says 'Nick Adams'" — although Nicholas Hart's circle is the better match for that row. He simply was not in `candidates`.

Folding the alias in before scoring (`if alias: candidates = candidates | {alias}`, after the `len(candidates) < 2` gate so the gate still keys on the first-token set) makes the comparison fair, and `settled` starts working for nickname aliases too — today a nickname alias can never reach `settled`, because `alias == top[1]` is unreachable when `alias` is not a candidate.

scripts/audit_aliases.py
```diff
@@ -172,33 +189,129 @@ def load_rows(path: Path) -> list[dict]:
...
-        for name, instr in parse_others(row.get("Others?") or row.get("Others") or ""):
-            cls = class_of(instr)
-            if name and cls:
-                people.append((expand_abbrev(name), cls))
-        for name, cls in people:
-            teammates = [n for n, _ in people if n != name]
-            appearances[(name, cls)].append(teammates)
+        people = row_people(row, slot_classes)
+        for name, cls, seat in people:
+            teammates = [n for n, _c, s2 in people if s2 != seat]
```
jsundram: **Seat-based exclusion makes a person written twice in one row their own teammate.**

The old line excluded by name (`n != name`); keying on `seat` means someone written in a player slot *and* in `Others?` — exactly the rows the dedup comment in `attribute_bare_entries` describes — now appears in their own teammate list. So `circles[X]` contains `X`, and `teammate_counter` can list someone as their own top collaborator in the variant report.

On `archive/data-raw.csv` as it stands that is 3 rows / 3 people, so it is small but live. Downstream it can flip an attribution: with `"Alice Hart" in circles["Alice Hart"]`, a row holding a bare `Alice` alongside a written-out `Alice Hart` scores 1 for Alice Hart and concludes the bare cell is the person already named in the same row.

Seat exclusion is the right call in `attribute_bare_entries` (the filled-row case the `row_people` docstring explains), but `collect_appearances` only ever sees rows as written, so it needs both:

```suggestion
            teammates = [n for n, _c, s2 in people if s2 != seat and n != name]
```

scripts/audit_aliases.py
```diff
@@ -226,7 +339,165 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+            by_first[(base_token(name), cls)].add(name)
+            circles[name].update(teammate_counter(apps))
+            written[name] += len(apps)
+    return by_first, circles, written
+
+
+def attribute_bare_entries(pairs: list[tuple[dict, dict]],
+                           appearances: dict[tuple[str, str], list[list[str]]],
+                           slot_classes: dict[str, str],
+                           ) -> tuple[list, list, list, int, int]:
```
jsundram: The annotation promises five elements, but the function returns six (`conflicts, unaliased, unsettled, unverified, settled, resolved_by_sheet`) and the docstring right below documents all six. Anyone unpacking from the signature rather than from the `return` gets a `ValueError`.

```suggestion
                           ) -> tuple[list, list, list, int, int, int]:
```

scripts/audit_aliases.py
```diff
@@ -308,12 +585,62 @@ def report_ambiguity(appearances: dict[tuple[str, str], list[list[str]]]) -> Non
...
+            print(f"   {'':10s} could not rule out: {', '.join(unruled)}"
+                  " (never written out)")
+    print(f"\n-- bare entries the room resolves but no alias covers ({len(unaliased)}) --")
+    print("   The app counts the bare form as its own person in every people")
+    print("   stat. The room already names them, so this is mechanical: write")
+    print("   the full name into the cell (or add the alias, if it is unambiguous).")
+    if not unaliased:
+        print("   (none)")
+    for row, name, cls, winner, why, unruled in unaliased:
+        print(f"   {describe_row(row)}  {name!r} [{cls}] -> {winner!r}"
```
jsundram: `cls` is `None` for an unannotated `Others?` subject — the `cls is None` branch in `attribute_bare_entries` exists for exactly those — so this prints a literal `[None]` where every other line in the report prints `[upper]`/`[cello]`. Verified on this branch with a bare `Alice` in an unannotated `Others?` cell:

```
   9/1/2024 Haydn          76#1            'Alice' [None] -> 'Alice Hart'  (played with Beryl, Carol)
```

Same on the `unsettled` line (631). `[{cls or 'any'}]` (or `'?'`) reads as the "no class, every namesake is in play" case the branch actually means.

scripts/audit_aliases.py
```diff
@@ -308,12 +585,62 @@ def report_ambiguity(appearances: dict[tuple[str, str], list[list[str]]]) -> Non
...
+              f"  (played with {', '.join(why[:3])})")
+        if unruled:
```
jsundram, line 687: `unruled` is built with `written[c] < MIN_WRITTEN_IN_FULL`, i.e. rivals written out 0-4 times, but the label says "(never written out)". A rival named four times in full — and possibly one with circle matches of its own, since `scored[1:]` is not filtered on score — is reported to the reader as never named. That understates the live alternative on the two lines whose whole job is "confirm before editing the sheet" (same wording at line 606). "(named fewer than 5 times)" or similar would match what was actually measured.

scripts/audit_all.sh
```diff
@@ -78,18 +78,50 @@ rule "DROPPED BY FILL-FORWARD   scripts/audit_fillforward.py"
...
+
+# The bare-entry findings the summary counts, printed from the raw pass so the
+# rows are reachable. Everything else in that pass duplicates the run above.
+rule "BARE NAMES        scripts/audit_aliases.py archive/data-raw.csv"
+sed -n '/^-- bare entries whose alias contradicts the room/,/no alias was consulted/p' \
+    "$RAW_OUT" 2>/dev/null || true
+
+rule "SUMMARY"
+printf '  %-46s %s\n' \
+  "bare names with 2+ candidates (for context)" "$(count 'bare names in the sheet with 2\+ candidates' "$OUT/aliases.txt")" \
```
jsundram: **The one hazard-1 number left in the summary still comes from the contaminated pass.**

Everything else about bare-name ambiguity moved to the raw pass in this PR, on the argument that on `archive/data.csv` `normalizePlayerNames` has already replaced each bare form with the alias's guess. That argument applies to this count too: a bare name that an alias covers is *gone* from the processed export, so the per-name list there is systematically short.

Verified against the real log on this branch:

```
archive/data.csv      -- bare names in the sheet with 2+ candidates (8) --
archive/data-raw.csv  -- bare names in the sheet with 2+ candidates (10) --
```

The summary prints 8, one line above three counts taken from the run that says 10. `$RAW_OUT` is already on disk here, so this can just use `n_of` like its neighbours:

```suggestion
  "bare names with 2+ candidates (for context)" "$(n_of 'bare names in the sheet with 2\+ candidates')" \
```

(`n_of` is defined two lines below; it would need moving above this `printf`.)

scripts/audit_all.sh
```diff
@@ -78,18 +78,50 @@ rule "DROPPED BY FILL-FORWARD   scripts/audit_fillforward.py"
...
+# throwing the body away would leave a reader with a number and no way to reach
+# the rows behind it. The processed pass prints its own attribution under a
+# "not evidence" banner, so it is no substitute.
+#
+# Warnings go to STDOUT, beside the numbers they qualify. fill_forward prints
+# its degradation notice on stdout for that reason, and re-emitting on stderr
+# would undo it: `npm run audit > report.txt` is the documented workflow, and a
+# degraded run must not read as a clean one in the file.
+RAW_OUT="$OUT/aliases-raw.txt"
+if ! "${PY[@]}" scripts/audit_aliases.py archive/data-raw.csv > "$RAW_OUT" 2>"$OUT/raw.err"; then
```
jsundram, line 95: **`raw.err` is only surfaced when the process exits non-zero, but the audit's other degradation is a stderr warning on an exit-0 run.**

`slot_annotation_classes()` prints `Note: could not class slot annotations via node (...); falling back to the slot position.` to **stderr** and returns `{}` — the script exits 0. When that happens every annotated slot reverts to positional classing, which changes `row_people`'s classes, which changes the class-keyed candidate sets in `candidate_index`, which changes the conflict/unaliased/unsettled verdicts printed in the BARE NAMES section below. The reader sees a clean report with silently different findings.

That is exactly the failure the fill-forward notice was moved to stdout to avoid (see the comment block just above, and `fill_forward`'s own comment). Either route this one to stdout too, or always echo `raw.err` when it is non-empty:

```bash
if ! "${PY[@]}" scripts/audit_aliases.py archive/data-raw.csv > "$RAW_OUT" 2>"$OUT/raw.err"; then
    echo "  !! the raw-sheet pass FAILED — the counts below are missing, not zero:"
fi
if [ -s "$OUT/raw.err" ]; then sed 's/^/     /' "$OUT/raw.err"; fi
```

scripts/audit_aliases.py
```diff
@@ -258,9 +557,7 @@ def report_ambiguity(appearances: dict[tuple[str, str], list[list[str]]]) -> Non
     for (name, cls), apps in appearances.items():
         present.add(name)
         resolved.add(EXISTING_ALIASES.get(name, {}).get(cls) or name)
-        if len(name.split()) > 1:
-            full_by_first[base_token(name)].add(name)
-        else:
+        if len(name.split()) == 1:
```
jsundram, line 603: **The per-name list and the per-entry list *can* disagree about the same name in the same output** — the thing `candidate_index`'s docstring says the shared definition prevents.

`bare_count` is built from `appearances`, and `collect_appearances` only keys an entry when `cls is not None` (line 319). An unannotated `Others?` name therefore never reaches `bare_count`, so it can never appear in "bare names in the sheet with 2+ candidates" — while `attribute_bare_entries` reads `row_people` directly, keeps `cls is None`, and reports it under `[any]`.

Reproduced on this branch (empty alias table, 5 attested rows each for `Alice Hart`/`Alice Bek`, plus one row with `Others? = "Alice"`):

```
-- bare names in the sheet with 2+ candidates (0) --
   (none)

-- bare entries the room resolves but no alias covers (1) --
   6/1/2024 Haydn  76#1  'Alice' [any] -> 'Alice Hart'  (played with Beryl)
```

The first number is what `audit_all.sh` prints in the SUMMARY, so an ambiguous name logged only in `Others?` without an instrument is invisible there. Counting cls-`None` entries into `bare_count` (under a synthetic `any` class, matching how they print) would close the gap.

scripts/audit_aliases.py
```diff
@@ -374,8 +724,18 @@ def main() -> None:
         print(f"CSV not found: {csv_path}", file=sys.stderr)
         sys.exit(1)
 
-    rows = load_rows(csv_path)
-    appearances = collect_appearances(rows)
+    # (row as written, row after fill-forward) — see fill_forward.
+    pairs = fill_forward(load_rows(csv_path))
```
jsundram, line 771: **Every section of the report is now sourced from the node round-trip, which silently drops rows.**

`fill_forward`'s bridge runs `dp.prepareRows(...)`, which filters out any row whose `Timestamp` doesn't parse as a JS `Date`, and the `{ dropped }` count it returns is thrown away. `main` then does `rows = [written for written, _filled in pairs]`, so those rows disappear from the variant grouping, the teammate counts, `bare_count`, `present`/`resolved`, and the canonicalized-export detector — not just from the attribution that needed the filled view.

Before this PR `load_rows` kept every row with a non-empty `Timestamp`. In a *data-quality* audit a malformed timestamp is itself a defect, and the printed `Rows: N` now silently disagrees with the file's row count with nothing to explain the gap. Returning `dropped` from the bridge and printing it (or at least warning when `len(pairs) != len(load_rows(...))`) would keep the loss visible.

scripts/run_audit_tests.sh
```diff
@@ -0,0 +1,42 @@
...
+        pytest test/test_audits.py "$@"
+fi
+
+# No uv: build a venv and install the pinned pytest into it. NOT
+# `pip install` into the system interpreter — that is PEP-668 marked on
+# current Ubuntu, so it would fail the CI step outright, and mutating the
+# runner's Python to run a test suite is the wrong trade anyway. NOT
+# "use whatever pytest is already importable" either: that is what the
+# previous version did, and it made the pin above fictional — CI ran the
+# image's pytest and only happened to match. The venv is cached between runs.
```
jsundram: **"The venv is cached between runs" is not true in CI**, which is the environment this fallback exists for (`uv` is not on `ubuntu-latest`).

`test.yml`'s `test` job caches only npm (`actions/setup-node` with `cache: 'npm'`); there is no `actions/cache` step for `$XDG_CACHE_HOME`/`~/.cache`, so every PR run does `python3 -m venv` + `pip install pytest==9.1.1` from PyPI. That is a per-run network dependency inside a job with `timeout-minutes: 10`, and a PyPI blip now fails an otherwise-green PR.

Either add a cache step keyed on `pytest-$PYTEST_VERSION`, or install `uv` in the workflow (`astral-sh/setup-uv`, pinned like the other actions) so CI takes the fast path the script prefers — and drop the caching claim from this comment.

scripts/audit_all.sh
```diff
@@ -78,18 +78,57 @@ rule "DROPPED BY FILL-FORWARD   scripts/audit_fillforward.py"
...
```
jsundram, line 95: **Architectural follow-up (2 of 3): make the data view a parameter, not a convention.**

This is the recurring bug class in this PR, and it is worth fixing structurally rather than one site at a time.

There are three views of the same data — raw, raw+fill-forwarded, and processed — and which one a section reads *determines whether it is correct*. Right now the rule lives entirely in prose comments, and the choice is expressed by picking a **file path in shell**: line 67 runs the audit on `archive/data.csv`, line 95 runs it again on `archive/data-raw.csv`. Python never sees which view it got, so it cannot check, and neither can a reader.

The cost is visible in this PR's history. Every review round found at least one section reading the wrong view:

- round 1: attribution built its circles from the processed export, so a wrong alias fed its own mis-attributed rows into the winner's circle
- round 2: switching to raw lost the `fillForward` evidence, so answerable entries were filed as NEEDS MEMORY
- round 4: `main()` fed *filled* rows to `collect_appearances`, so `bare_count`/`present`/`sheet_surnames` all read synthesised values
- round 6: the summary's "bare names with 2+ candidates" still reads `$OUT/aliases.txt` (processed, 8) while its three new neighbours read the raw pass (10)

That last one is the same bug as round 1, relocated. The tell is not that bugs exist — it is that the identical class of bug keeps reappearing somewhere new, which means nothing structural prevents it.

`fill_forward()` already has the right idea: it returns `(written, filled)` pairs precisely because the two answer different questions. The fix is to finish that thought:

- one loader produces the view set once, and every section takes the pair (or a named view), never a bare `list[dict]`
- `collect_appearances`, `candidate_index` and `report_ambiguity` declare which view they need in their signature, so a caller cannot hand over the wrong one silently
- `audit_all.sh` stops choosing views at all — it invokes the audit once, and the report decides internally which view each section reads

Second-order benefit: the "a 0 here is not evidence" banner and the CANONICALIZED caveat both exist to warn a reader that a section is reading a view it should not. Once the view is structural, those warnings have nothing left to warn about.

Related: #21 covers the other half of this (the Python/JS seam). This item stands on its own and is worth doing whether or not that port happens.

scripts/audit_aliases.py
```diff
@@ -226,7 +357,189 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+    if cls is None or cls == ANY_CLASS:
+        return {c for (tok, _cls), names in by_first.items() if tok == token
+                for c in names}
+    return by_first.get((token, cls), set()) | by_first.get((token, ANY_CLASS), set())
+
+
+def attribute_bare_entries(pairs: list[tuple[dict, dict]],
+                           appearances: dict[tuple[str, str], list[list[str]]],
+                           slot_classes: dict[str, str],
+                           ) -> tuple[list, list, list, int, int, int]:
```
jsundram, line 434: **Architectural follow-up (3 of 3): split attribution from the descriptive audit, and collapse the buckets to the actions available.**

This script used to be descriptive — count variants, group by first token, print. It is now an inferential engine: scoring, `MIN_WRITTEN_IN_FULL`, tie-breaking, unruled-rival disclosure, and a six-element return of five outcome buckets plus a counter. That is a change in *kind*, made in place, and the report's vocabulary has not caught up with it.

Look at what the review findings across this PR actually were. Very few were wrong arithmetic. Nearly all were the report's **label** disagreeing with its **computation**:

- `settled` meant both "the room agrees with an existing alias" and "no alias exists at all", and both printed as "agree with the table — nothing to do"
- `settled` was later also incremented for entries the *sheet* resolved via fill-forward, where no alias was consulted at all — crediting `src/aliases.js` with fill-forward's work
- "(never written out)" labelled `unruled`, which is `written[c] < MIN_WRITTEN_IN_FULL`, i.e. 0–4 appearances — on the very lines asking the reader to confirm before editing the sheet
- an unannotated `Others?` subject printed a literal `[None]`
- the per-name and per-entry sections could disagree about the same name in one run, which `candidate_index`'s docstring says is impossible
- counts were per person-occurrence while every label said "bare rows"

Six findings, one cause: the buckets are named after *internal states of the algorithm* rather than after what the reader should do. Five buckets do not map onto five distinct actions, so every bucket boundary needs a paragraph of prose to explain, and each paragraph is another chance for the words and the code to drift apart.

There are exactly three things a reader can do with a finding:

1. **Edit this sheet cell** — the room contradicts the alias, or names someone no alias covers (today: `conflicts` + `unaliased`)
2. **Add this alias** — a bare form the room resolves confidently and that recurs
3. **Ask someone, now, while they still remember** — no evidence and no alias (today: `unsettled`)

Everything else — `unverified`, `settled`, `resolved_by_sheet` — is not a finding. It is a count that says "the tool looked and found no work here," and it belongs in the summary as a single reassurance line, not as a bucket with its own semantics to explain and mislabel.

Concretely:

- return a list of findings, each carrying its action, rather than a 6-tuple of parallel lists (which also removes the annotation/arity mismatches this PR has already had twice)
- move attribution into its own module with its own entry point. It has a different lifecycle from the descriptive audit — CLAUDE.md already notes the alias findings decay fast and the rest can batch, so they want different run cadences and shouldn't share one 881-line file and one 200-line printer
- let the descriptive audit go back to describing

This is the change most likely to reduce review churn on the next round, because it removes the surface where the churn has been happening.

scripts/audit_aliases.py
```diff
@@ -172,33 +195,141 @@ def load_rows(path: Path) -> list[dict]:
...
+        people = row_people(row, slot_classes)
+        for name, cls, seat in people:
+            # Both tests are needed. The seat drops this very cell; the name
+            # drops the SAME person written again elsewhere in the row — a
+            # player slot and an Others? entry, which is how the rows that
+            # overflow the quartet layout get logged. Without it circles[X]
+            # contains X, and a bare name sitting beside its own full form
+            # scores a point for being the person already named in that row.
+            teammates = [n for n, _c, s2 in people if s2 != seat and n != name]
+            appearances[(name, cls if cls is not None else ANY_CLASS)].append(
```
jsundram, line 331: **`ANY_CLASS` leaks into the paste-ready `PLAYER_ALIASES` proposal, producing entries the app can never use.**

Bucketing instrument-less `Others?` entries under `"any"` puts that pseudo-class into `appearances`, and `main()` feeds `appearances` straight into the variant grouper (`groups` -> `by_class` -> `proposals`, line ~809) and the REVIEW table. Neither was taught that `"any"` is not a real class, so the audit now proposes aliases keyed on it.

Reproduced on a fixture (placeholder names, unannotated `Others?` cells only):

```
=== 'zoe' (2 variants) ===
  'Zoe Hart'   [any  ]  3x   teammates: Bob Smith x3, ...
  'Zoe'        [any  ]  1x   teammates: Bob Smith x1, ...
    -> propose 'Zoe' [any] -> 'Zoe Hart'  (overlap=100%, 1x)
...
=== PLAYER_ALIASES proposal (paste into src/aliases.js ...) ===
    "Zoe": { any: "Zoe Hart" },
```

Concretely, this misbehaves three ways:

1. `canonicalize(name, cls, aliases)` only ever looks up `cls` of `'upper'`/`'cello'` (`classOf(o.instrument)` is `null` for an unannotated `Others?` entry), so a pasted `{ any: ... }` entry never fires. The user is told to paste something inert.
2. Once pasted it becomes actively misleading: `report_ambiguity` section 1 reads `EXISTING_ALIASES.get(name, {}).get(cls)` with `cls == "any"`, so the very next run prints `'Zoe' [any] ... alias says 'Zoe Hart'` — the name now reads as handled while the app still counts the bare form as its own person.
3. Until it is pasted, `already_aliased(variant, "any")` can never be true, so the same dead proposal is re-emitted on every single run even after the user adds a correct `upper`/`cello` entry for that key.

The `unaliased` bucket's advice ("or add the alias, if it is unambiguous") has the same problem for `[any]` entries — no alias can reach them.

Suggest excluding `ANY_CLASS` from the proposal/REVIEW paths (it is only meant to feed the per-name list and the candidate sets), e.g. skip `cls == ANY_CLASS` in the two `for cls, vs in by_class.items():` loops in `main()`.

scripts/audit_aliases.py
```diff
@@ -226,7 +357,189 @@ def already_aliased(variant: str, cls: str) -> bool:
...
+    the population the threshold exists to protect against.
+    """
+    by_first: dict[tuple[str, str], set[str]] = defaultdict(set)
+    circles: dict[str, set[str]] = defaultdict(set)
+    written: Counter = Counter()
+    for (name, cls), apps in appearances.items():
+        tokens = name.split()
+        if len(tokens) > 1 and len(tokens[-1].rstrip(".")) > 1:
+            by_first[(base_token(name), cls)].add(name)
+            circles[name].update(teammate_counter(apps))
```
jsundram: **`circles` is built from the rows *as written* while `mates` comes from the *filled* row — the asymmetry re-creates the false NEEDS MEMORY this feature exists to remove.**

`written[name]` genuinely has to come from the written rows (the docstring's argument about a name typed once clearing `MIN_WRITTEN_IN_FULL` five times over is right). But `circles[name]` is a different question: a teammate established by fill-forward is a real fact about who was in the room, not the sheet repeating itself. On the raw sheet — which this section *must* be run against — a continuation row has no players written at all, so any full name typed on such a row gets an empty teammate list, and its circle stays empty no matter how many sessions it played.

Reproduced on a fixture, exactly the raw-sheet convention the PR is built for (session states the quartet on row 1, names the extra player in `Others?` on row 2):

```
1/1/2024 10:00, Bob Smith, Carol Jones, Dan Ray, (others empty)
1/1/2024 11:00, (blank)  , (blank)     , (blank), Zoe Hart
   ... x6 sessions, so written["Zoe Hart"] == 6 >= MIN_WRITTEN_IN_FULL
12/1/2024 10:00, Zoe, Bob Smith, Carol Jones
```

`Zoe Hart` played with Bob Smith and Carol Jones six times, and the bare-`Zoe` row names both of them, yet:

```
-- bare entries the room resolves but no alias covers (0) --
   (none)
-- bare entries nobody has decided (1) --
    12/1/2024 Haydn  64#1   'Zoe' [upper]
              candidates: Zoe Bek, Zoe Hart
```

`circles["Zoe Hart"]` is empty, `top[0] == 0`, and the entry is reported as needing memory. Building `circles` from the filled view (while keeping `written` on the written view) would settle it — that is the split `fill_forward` already returns both rows for.

scripts/audit_all.sh
```diff
@@ -78,18 +78,57 @@ rule "DROPPED BY FILL-FORWARD   scripts/audit_fillforward.py"
...
+if [ -s "$OUT/raw.err" ]; then sed 's/^/     /' "$OUT/raw.err"; fi
+RAWALIAS=$(cat "$RAW_OUT" 2>/dev/null || true)
+if printf '%s' "$RAWALIAS" | grep -q 'could not fill-forward'; then
+    echo "  !! fill-forward degraded on the raw pass — NEEDS MEMORY below is inflated."
+fi
+
+# The bare-entry findings the summary counts, printed from the raw pass so the
+# rows are reachable. Everything else in that pass duplicates the run above.
+rule "BARE NAMES        scripts/audit_aliases.py archive/data-raw.csv"
+sed -n '/^-- bare entries whose alias contradicts the room/,/no alias was consulted/p' \
```
jsundram: **The SUMMARY's "bare names with 2+ candidates" count now comes from the raw pass, but the only per-name list a reader can see in the output is the processed pass's — so the number and the list disagree with no way to reach the difference.**

Line 119 switched that count from `count(... "$OUT/aliases.txt")` (processed) to `n_of` (raw), and the comment right above it notes the two differ ("10 names there against 12 raw"). The `BARE NAMES` block printed here starts at `^-- bare entries whose alias contradicts the room`, so the raw pass's `-- bare names in the sheet with 2+ candidates --` section is never shown; the only such list in the report is the processed one printed at line 67. A reader sees `12` in the summary, counts 10 entries in the body, and the two extra names — precisely the ones an alias is currently covering, i.e. the interesting ones — are unreachable.

Starting the range one section earlier fixes it:

```suggestion
sed -n '/^-- bare names in the sheet with 2+ candidates/,/no alias was consulted/p' \
    "$RAW_OUT" 2>/dev/null || true
```

.github/workflows/test.yml
```diff
@@ -32,6 +32,22 @@ jobs:
...
         run: npm test
 
+      # The audits are Python and node:test cannot see them; this is where
+      # every bug in the alias/ensemble tooling has surfaced, so it gates too.
+      # No new action to pin: the uv-or-venv fallback lives in the script.
+      #
+      # ubuntu-latest ships no uv, so the venv path runs and pip fetches the
+      # pinned pytest. Cached so that is a one-time cost rather than a PyPI
+      # round-trip on every PR, inside a job with a 10-minute timeout.
+      - name: Cache pip downloads
```
jsundram, line 45: **This cache will almost never hit, so `pip install pytest` does hit PyPI on essentially every PR — contrary to both this comment and `run_audit_tests.sh`'s "the install does not depend on PyPI being reachable".**

`test.yml` triggers only on `pull_request` (and `workflow_dispatch`), never on push to `main`. GitHub Actions caches written by a PR run are scoped to that PR's ref and can only be restored by the same PR (or by runs on the branch that created them); other branches can only read caches created on the *default* branch. Since nothing in this repo ever populates `pip-pytest-*` on `main`, every new PR starts cold and downloads pytest from PyPI. Only repeat runs *within one PR* benefit.

There is also no `restore-keys`, so any edit to `scripts/run_audit_tests.sh` (which includes bumping `PYTEST_VERSION`, the one change where a near-miss restore would still help) invalidates even that.

If cross-PR reuse is wanted, the cache has to be seeded by a workflow that runs on `main` — otherwise the comment is worth softening so a future reader does not rely on a guarantee that is not there.

jsundram:
Retired rather than merged. Most of this diff is deleted or rewritten by the work it prompted, so landing it would mean debugging code we have already decided to replace.

The branch is archived as the tag `archive/attribution-spike` (a7f87b2) — the only executable record of the attribution heuristic and the rules six review rounds established, including `test/test_audits.py`'s 61 cases. A rewrite plan and derived issues follow; the tag can be deleted once those are closed.

What this branch established, and what carries forward:
- the heuristic works — a row's other players identify a bare first name, and the finding count stayed stable at 2 through every round while everything around it churned
- `class_of` had drifted from the app's `classOf` on the assistant prefix and on spelled-out cello
- the recurring bug classes are structural: a Python/JS seam, data views chosen by convention, and buckets named after algorithm states

PR #20, opened 2026-08-28
A blank cell is a ditto mark, whatever the gap

jsundram:
Split out of #19, which it had grown into. This one edits the data layer and **moves user-visible stats**; #19 only reads the sheet.

> *"I would never enter a session with blanks; if there are blanks it is because I'm expecting fillForward to fix them for me."*

That settles what a blank means, and `SESSION_WINDOW_HOURS` had no business gating it. A blank can only mean "same as above" — leaving the names out is not how anyone starts a group — and `-` already exists to say "nobody in this seat", which is what keeps the two distinguishable.

Time-gating it was not merely conservative, it was destructive. A row outside the window fell through to the branch that assigns the entry it was given, an empty string, and *that* became the anchor for every row after it:

```
8/3/2024 15:35  Haydn 50#2     Isaac, Elaine, Shay
8/3/2024 16:22  Haydn 54#2     (blank — fills fine)
8/3/2024 20:29  Beethoven 18#6  ← 4.11h dinner break; falls outside
8/3/2024 21:20  Mozart K589     ← inherits the blank
8/3/2024 22:22  Brahms 1        ← inherits the blank
```

An afternoon session, a dinner break, an evening one. All three evening rows had **no players at all**. Six rows in the log did; now zero do.

**This changes stats.** Those six rows gain three players each, so unique-people counts, collaborator totals and network edges all move — toward counting people who were actually there.

## What stays gated

The written-shorthand rule, because that one is an inference rather than a ditto: `Peter` an hour after `Peter Ouyang` abbreviates it; months later it is as likely to be a different Peter.

## The window is now derived, not asserted

Since it governs exactly one rule, `audit_fillforward` measures that one — the gaps at which shorthand is actually typed:

```
9 shorthand entries; 8 inside the window.
    0.04h ... 2.54h   [expanded]
   13.13h            [left as typed]
```

Any value between 2.6h and 13h behaves identically, which is a far better answer about 4h than a continuation-gap distribution was. The `Others?`-drop check drops the window too, for the same reason the app does: a blank row inherits its anchor whatever the gap, so the rows it used to skip are the ones worth reporting.

## Docs

- `md/howto.md` §6 is new — what repeats and what doesn't, for someone mid-entry: repetition is per field so a rotating stand only needs the seat that changed; `Others?` never repeats; a blank always does; `-` means nobody.
- `CLAUDE.md` records why `fillForward` must run before `normalizePlayerNames`. Three things break if swapped — the prefix rule losing to a global guess, the abbreviation chain stalling halfway, annotations not riding along on a filled slot — each tested rather than asserted.

Two node tests that pinned the old behaviour are inverted to pin the new one. 320 node tests, 93 pytest, lint and typecheck clean.

**Based on `attribute-bare-rows`** so this diff shows only the fillForward change; retarget to `main` once #19 lands.


jsundram:
Superseded by a branch cut from `main`. This one was stacked on #19, which is being retired, so its diff could not stand alone.

PR #24, opened 2026-08-29
A blank cell is a ditto mark, whatever the gap

jsundram:
Cut from `main`, independent of the audit tooling work. Replaces #20, which was stacked on #19.

> *"I would never enter a session with blanks; if there are blanks it is because I'm expecting fillForward to fix them for me."*

That settles what a blank means, and `SESSION_WINDOW_HOURS` had no business gating it. A blank can only mean "same as above" — leaving the names out is not how anyone starts a group — and `-` already exists to say "nobody in this seat", which is what keeps the two distinguishable.

Time-gating it was not merely conservative, it was destructive. A row outside the window fell through to the branch that assigns the entry it was given, an empty string, and *that* became the anchor for every row after it:

```
8/3/2024 15:35  Haydn 50#2      Isaac, Elaine, Shay
8/3/2024 16:22  Haydn 54#2      (blank — fills fine)
8/3/2024 20:29  Beethoven 18#6  ← 4.11h dinner break; falls outside
8/3/2024 21:20  Mozart K589     ← inherits the blank
8/3/2024 22:22  Brahms 1        ← inherits the blank
```

An afternoon session, a dinner break, an evening one. All three evening rows had **no players at all**. Six rows in the log did; now zero do.

**This changes stats.** Those six rows gain three players each, so unique-people counts, collaborator totals and network edges all move — toward counting people who were actually there.

## What stays gated

The written-shorthand rule, because that one is an inference rather than a ditto: `Peter` an hour after `Peter Ouyang` abbreviates it; months later it is as likely to be a different Peter.

## The window is now derived, not asserted

Since it governs exactly one rule, `audit_fillforward` measures that one — the gaps at which shorthand is actually typed:

```
9 shorthand entries; 8 inside the window.
    0.04h ... 2.54h   [expanded]
   13.13h            [left as typed]
```

Any value between 2.6h and 13h behaves identically. The `Others?`-drop check drops the window too, for the same reason the app does: a blank row inherits its anchor whatever the gap.

## Docs

- `md/howto.md` §6 is new — what repeats and what doesn't, for someone mid-entry: repetition is per field so a rotating stand only needs the seat that changed; `Others?` never repeats; a blank always does; `-` means nobody.
- `CLAUDE.md` records why `fillForward` must run before `normalizePlayerNames` — three things break if swapped, each tested rather than asserted.

Two tests that pinned the old behaviour are inverted to pin the new one. 320 tests, lint and typecheck clean.


jsundram:
Merging with one known gap: `session_window_report` ships untested. The three cases exist in `archive/attribution-spike` and are tracked as test debt on #26, which is where they land when the audits move to node:test. The behaviour change itself — a blank cell filling regardless of the gap — is covered by the two inverted tests in `test/dataProcessor.test.mjs`.

PR #29, opened 2026-08-31
Add a 1M date range option and fix the range boundary math

jsundram:
Adds a `1M` button to the date range control and fixes three boundary bugs the other ranges shared.

The strip now reads **All / YTD / 1Y / 6M / 1M / Custom**, in descending window length. `DateFilterWidget` is mounted twice — Home (`#dateSlider`, via `NavigationComponent`) and Dashboard (`#dashboardDateFilter`) — and the range list lives inside it, so both screens get the option identically; nothing outside the widget enumerates the range ids.

## Clamping the month arithmetic

`1M` starts at the same day of the previous month, falling back to that month's last day when the previous month is shorter (Mar 31 → Feb 29 in a leap year, Feb 28 otherwise).

The obvious `setMonth(getMonth() - 1)` **overflows rather than clamps**: Mar 31 becomes Feb 31, which `Date` rolls forward to Mar 2/3 — putting the start of the window *inside the month after* the one it should cover. `monthsAgo(date, months)` shifts from the 1st and then sets the day to `min(day, lastDayOfTargetMonth)`.

`6M` and `1Y` had exactly this bug and now use the same helper:

| range | before | after |
| --- | --- | --- |
| `6M` on Aug 31 | Mar 3 (window starts a month late) | Feb 28 |
| `1Y` on Feb 29 | Mar 1 of the prior year (364-day window) | Feb 28 |

## Whole days at both ends

Starts carried the **current clock time** onto the boundary day, so `1M` chosen at 14:00 on Aug 31 began at Jul 31 14:00 and dropped a session logged that morning — while `Custom` with the same two visible dates included it, since `fromDateInputValue` anchors at midnight. `startOfDay()` is applied to whatever the switch produces, so every preset is anchored by one rule rather than each branch having its own.

The **end** had the same staleness, and it bit harder ([thread](https://github.com/jsundram/quartet-log/pull/29#discussion_r3896074594)). The bounds were computed only in the constructor and on a range click, so the end froze at whenever the button was last pressed — and `filterEngine.filterRows` is inclusive against it. A session logged after that moment was fetched by `revalidate()` and then filtered straight back out; an installed PWA resumed the next morning showed "1M as of yesterday" until you clicked a range button.

So relative windows are no longer cached at all: `getRange()` derives them per read, and `1M` means "the last month" rather than "the month before the click". `CUSTOM` keeps its explicit pair. Both consumers already called `getRange()` on every render, so the change is contained.

## Where the arithmetic lives

`monthsAgo`, `startOfDay` and the range switch moved to **`src/dateRange.js`**, which imports nothing — the `ALL` branch takes `getBeginDate` as an argument rather than importing `getBegin`, since that throws until `setBegin` has seen data and `App` constructs the widget (line 31) well before that (line 119).

That keeps the calendar tests off `d3` and the gitignored alias table, and it makes the range switch itself testable — which mattered, because the switch is where both clamping bugs actually lived and nothing covered it ([thread](https://github.com/jsundram/quartet-log/pull/29#discussion_r3896075882)). The widget is now loaded by one test file, for the wiring only it can show.

## Verification

`npm test` (341 pass), `npm run lint`, `npm run typecheck` all clean.

Mutation-checked the new coverage rather than assuming it bites — `case '1M'` → `monthsAgo(now, 11)` fails 2 tests, dropping `startOfDay` fails 2, and reverting `getRange` to the cached pair fails 1.

Also driven in headless Chromium with a pinned clock: 6 buttons in order, every range landing on the right day across leap years, common years, short target months and year boundaries, every start reading `00:00:00`, the end advancing between reads without re-clicking, and the `CUSTOM` pair surviving untouched. Separately measured the strip against the real stylesheet at 320–430px — 252px wide, `Custom` fully visible, no overflow at any phone width ([thread](https://github.com/jsundram/quartet-log/pull/29#discussion_r3896077163)).

`./build.sh --prod` isn't runnable in my container (no pandoc), so CI's `smoke` job is the first full-build check.

https://claude.ai/code/session_014NnRZJAKARGP7SPH9wCdqm


src/dateFilterWidget.js
```diff
@@ -121,18 +168,19 @@ export class DateFilterWidget {
...
+            case '1M':
+                start = monthsAgo(now, 1);
                 break;
             default:
                 start = getBegin();
         }
 
-        this.startDate = start;
+        this.startDate = startOfDay(start);
         this.endDate = now;
```
jsundram: The *end* of the window has the same class of bug this PR fixes at the start, and it bites harder with `1M`.

`updateDatesFromRange` runs in exactly two places — the constructor and `handleRangeClick` — so `this.endDate` is frozen at whatever `new Date()` was when the range was last chosen. `filterRows` is inclusive on both sides (`start <= d.timestamp && d.timestamp <= end`, `src/filterEngine.js:21`), so any row whose timestamp is later than that frozen moment is filtered out.

That isn't hypothetical here: `revalidate()` is called on foreground resume, on the 5-minute poll, and on pull-to-refresh. A session logged an hour after the tab was opened gets fetched, gets re-rendered into `filteredData`… and then fails `d.timestamp <= end`. In an installed PWA resumed the next morning, "1M" is silently "1M as of yesterday" and today's sessions are invisible until the user clicks a range button.

The PR description addresses this and I think it lands on the wrong reason:

> The end of the window stays at `now`. Sheet timestamps are submission times, so nothing can be logged ahead of the clock and there is no equivalent gap at that end.

That rules out *future* timestamps, but the gap isn't ahead of the clock — it's between the cached `now` and the real one. Nothing can be logged ahead of the clock, but plenty gets logged ahead of `this.endDate`.

The fix that matches the spirit of the rest of the PR is to stop caching a relative window. Either recompute in `getRange()` when `currentRange !== 'CUSTOM'`, or store the range id and derive the pair on read — so `1M` means "the last month" rather than "the month before the click". `CUSTOM` keeps its explicit pair. Happy to see it split out as a follow-up rather than grown into this PR, but it does undercut the headline claim that every preset now covers a whole number of days: the last day is still a partial one, just at the far end.

---

src/dateFilterWidget.js
```diff
@@ -121,18 +168,19 @@ export class DateFilterWidget {
...
+                break;
+            case '1M':
+                start = monthsAgo(now, 1);
                 break;
             default:
                 start = getBegin();
         }
 
-        this.startDate = start;
+        this.startDate = startOfDay(start);
```
jsundram: The two helpers are thoroughly tested; the switch that *uses* them is not, and that's where the bugs being fixed actually lived. Nothing in `test/dateFilterWidget.test.mjs` would fail if `case '1M'` were edited to `monthsAgo(now, 11)`, or if `startOfDay` were dropped from line 183 and reapplied inside `monthsAgo` only (leaving `ALL`/`YTD` unanchored).

This is cheap to pin down because `updateDatesFromRange` needs no DOM — `root()` is only called from `render()`/`handleRangeClick`, so the constructor runs fine under `node:test`:

```js
const w = new DateFilterWidget('#x', () => {}, { defaultRange: '1M' });
const [start] = w.getRange();
assert.equal(start.getHours(), 0);
assert.equal(ymd(start), ymd(monthsAgo(new Date(), 1)));
```

`ALL` needs `setBegin()` first (`getBegin()` throws otherwise), which is itself worth a case — it's the one branch where the new `startOfDay` wrapper is a pure no-op today, since `setBegin` already normalizes to the 1st of the month at midnight. Worth a comment there so a future reader doesn't take the wrapper as evidence that `getBegin()` returns an arbitrary time.

---

src/dateFilterWidget.js
```diff
@@ -33,6 +79,7 @@ export class DateFilterWidget {
             { id: 'YTD', label: 'YTD' },
             { id: '1Y', label: '1Y' },
             { id: '6M', label: '6M' },
+            { id: '1M', label: '1M' },
```
jsundram, line 42: Worth checking this strip at 320–390px before merge. `.date-filter-container` is `overflow-x: auto` and `.date-range-buttons` is `flex-shrink: 0`, so a sixth button doesn't wrap or squeeze — it pushes the group wider and the overflow scrolls. Rough arithmetic at the mobile padding (`7px 10px`) puts six buttons around 300px, with `Custom` the one that falls off the right edge. It's reachable by scrolling, but the least discoverable of the six ends up hidden by default on the smallest phones.

There's a second-order effect too, from the pull-to-refresh guard: `startsInScroller` disables PTR for any gesture starting inside an element that is declared scrollable **and actually overflows** on either axis. The date strip is already called out in CLAUDE.md as one of those. Today it only overflows on narrow screens; the sixth button widens the band of viewports where it does, so PTR goes dead over the strip on more devices. That's the documented accepted cost rather than a new bug, but it's a real consequence of the extra button and worth knowing you're paying it.

If `Custom` scrolling off does bother you, shortening the `All` label or trimming mobile padding a couple of px buys back the room without touching the layout model.

---

test/dateFilterWidget.test.mjs
```diff
@@ -0,0 +1,101 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+
+import { monthsAgo, startOfDay } from '../src/dateFilterWidget.js';
```
jsundram: Small structural note: importing `dateFilterWidget.js` pulls in its whole module graph, which is `d3` and `./config.js` → `./aliases.js`. The test never touches the name tables, so nothing here is *wrong* — `pretest` materializes the stub and the suite runs in well under a second either way. But it does mean this file loads the gitignored personal table into the test process on your machine and the empty stub in CI, which is the exact asymmetry CLAUDE.md went to some trouble to make structurally impossible for the alias helpers.

`monthsAgo` and `startOfDay` are pure calendar math with no widget dependency at all. Moving them to a small `src/dateUtils.js` (or into `dataProcessor.js`, which is deliberately import-free) would let this test import only what it exercises, and would make the "test the switch too" suggestion on `updateDatesFromRange` the only place the widget itself needs loading.

Not a merge blocker — flagging it because the convention elsewhere in the repo is strict enough that the exception is worth a deliberate decision rather than a default.

---

jsundram (commented):
The month arithmetic is right and the two clamping bugs it retires are real — I reproduced the old `6M`-on-Aug-31 → Mar 3 and `1Y`-on-Feb-29 → Mar 1 results, and `monthsAgo` fixes both by construction rather than by special case. Shifting from the 1st before the month arithmetic is the correct trick, and the test table covers the interesting shapes (leap/common, 31→30, 31→28, year crossing, non-mutation). Confining the range ids to the widget holds up: `grep` finds no `'6M'`/`'1Y'` outside `dateFilterWidget.js`, so `1M` really is a one-line addition for both mounts.

`npm test` (331 pass), `npm run lint` and `npm run typecheck` are all clean on `62ca497` here.

Four comments inline. The one I'd want resolved before merge is the window's **end**: `this.endDate` is frozen at the moment the range was last clicked, and `filterRows` is inclusive against it, so a session logged after page load is fetched by `revalidate()` and then filtered straight back out. That's the same "the boundary carries a stale clock time" failure this PR fixes at the start, and the description's argument for leaving it (submission times can't run ahead of the clock) doesn't cover it. Fixing it is a small change — derive the pair in `getRange()` for non-`CUSTOM` ranges — but a follow-up PR is fine if you'd rather keep this one tight; I'd just soften the claim in the description if so.

The rest are smaller: no test covers the `switch` wiring itself (only the helpers it calls), the six-button strip likely scrolls `Custom` off-screen at 320px, and the new test file drags `d3` and the gitignored alias table into its module graph for two pure date functions.

---

src/dateFilterWidget.js
jsundram: Confirmed and fixed in `19067f9` — you're right that the description's reason was the wrong one. The gap isn't ahead of the clock, it's between the cached `now` and the real one, and nothing about submission times closes it.

`getRange()` now calls `presetBounds(this.currentRange, new Date(), getBegin)` per read, so a relative window is derived from the clock at the moment it's used. `CUSTOM` short-circuits and returns the stored pair, since that one genuinely is explicit user input. The stored pair survives only to prefill the Custom inputs, which is now the only thing `updateDatesFromRange` is for — noted in a comment on it so the next reader doesn't take it for the source of truth.

Both consumers already call `getRange()` on every render (`filterEngine` via `App.filterData`, `DashboardComponent.filteredRows`), so nothing else had to change and the re-derivation costs a `Date` construction per filter pass.

Verified in a browser: click `1M`, wait, read again without touching the control — the end advanced 1101ms while the start held at Jul 31 00:00. There's a regression test for it too (`re-derives a preset window on every read`), and I checked it fails if `getRange` goes back to serving the cached pair.

I've dropped the paragraph making that argument from the PR description.

---

src/dateFilterWidget.js
jsundram: Agreed — the switch was the untested part and it's where both bugs lived. Covered in `19067f9`.

Moving the switch into `presetBounds` (see the reply on the test file) means most of it is now testable without loading the widget at all: `test/dateRange.test.mjs` pins each range id to a fixed expected start against a pinned `NOW`, asserts every start is midnight including `ALL`/`YTD`, and covers the unknown-id fallback. `test/dateFilterWidget.test.mjs` keeps the wiring only the widget can show — id → bounds, the CUSTOM short-circuit, and the per-read derivation.

I ran your two mutations to check the tests actually bite:

| mutation | result |
| --- | --- |
| `case '1M'` → `monthsAgo(now, 11)` | 2 failures |
| `startOfDay` dropped from the return | 2 failures |
| `getRange` serves the cached pair | 1 failure |

On `ALL`: `setBegin` normalizes to the 1st of the month at midnight, so the anchor is a no-op there — there's now a comment saying exactly that at the `case 'ALL'` branch, plus a test that calls `setBegin(new Date(2019, 4, 17, 9, 30))` and asserts `2019-05-01`. And rather than needing `setBegin` before the other cases, `presetBounds` takes `getBeginDate` as an argument and only the branches that need it call it — there's a test that passes a throwing stub and asserts `YTD`/`1Y`/`6M`/`1M` never touch it. That's load-bearing beyond tidiness: `App` constructs `NavigationComponent` (line 31) before `setBegin` (line 119), so an eager read would throw at construction.

---

test/dateFilterWidget.test.mjs
jsundram: Taken — done in `19067f9`, as a deliberate decision rather than a default.

`monthsAgo`, `startOfDay` and the range switch now live in `src/dateRange.js`, which imports nothing at all. It couldn't import `getBegin` even if it wanted to: the `ALL` branch takes it as an injected `getBeginDate` argument, because `getBegin()` throws until `setBegin` has seen data and `App` constructs the widget before that. So the same lazy-read requirement that the runtime has also keeps the module import-free — the two constraints happened to agree.

`test/dateRange.test.mjs` (the old file, renamed) now imports only that module: no `d3`, no `config.js` → `aliases.js`. `test/dateFilterWidget.test.mjs` is a new, smaller file that loads the widget for exactly the wiring you asked about on `updateDatesFromRange` — which is the one place you predicted would still need it.

I put it in its own module rather than `dataProcessor.js`: that file's import-free-ness is in service of the row-processing pipeline, and range-picker arithmetic isn't part of it. Happy to fold it in if you'd rather have one home for pure helpers.

---

src/dateFilterWidget.js
jsundram: Checked it rather than estimating — headless Chromium against the real `static/css/viz.css`, measuring the live boxes. It doesn't reproduce: **`Custom` is fully visible at 320px and the strip never overflows at any phone width.**

| viewport | group width | container | overflows | `Custom` right edge |
| --- | --- | --- | --- | --- |
| 320px | 252px | 320px | no | 256px |
| 360px | 252px | 360px | no | 256px |
| 390px | 252px | 390px | no | 256px |

Per-button at mobile: `All` 36 · `YTD` 46 · `1Y` 36 · `6M` 38 · `1M` 38 · `Custom` 57 = 252px, against ~310px of usable width.

The arithmetic came out ~50px under your estimate because of the font size, which I'd have missed too: `.date-range-btn` is `font: inherit`, and `@media (max-width: 800px)` sets `body { font-size: 12px }` (viz.css:1256). So the mobile strip renders at 12px, not the 16px base. The group was 215px before this PR and is 252px after; it only starts overflowing below a ~262px viewport, narrower than any shipping phone.

That also means the pull-to-refresh consequence doesn't materialize. `startsInScroller` bails only on an element that is declared scrollable **and actually overflows**, and the strip does neither at real widths — before or after the sixth button. It's `overflow-x: auto` on a box that isn't overflowing, which is the case CLAUDE.md explicitly calls out as staying pullable ("an `overflow-x` wrapper that doesn't currently overflow… stays pullable"). So no band of viewports widens and no cost is being paid here.

No change made, on the grounds that there's nothing to buy back. Shout if you want the headroom anyway for a future seventh button.

---

PR #30, opened 2026-08-31
Fit dashboard ranked-chart names to the left margin

jsundram:
Long musician names ran off the left edge of the dashboard's Top 20 chart. The row name is right-anchored 8px left of the bar and nothing bounded it, so anything wider than the margin (120px desktop, 96px mobile) was simply off-screen.

## What changed

Names are now fitted to the gutter, per row:

1. **Full name** when it fits — unchanged for most rows at desktop width.
2. **The short form** otherwise, reusing the `disambiguateLabels` logic the network views already use: `Aaron` when that first name is unique among the chart's musicians, `Aaron J.` when it isn't, full name when even that collides.
3. **An ellipsis clip** if nothing fits, so a label is never silently cut off by the viewport edge.

A shortened label carries the full name in a `<title>` for hover.

## Implementation notes

- `fitText(candidates, maxWidth, measure)` (`src/dataProcessor.js`) is pure and takes the measurer as an argument, so it's unit-testable. The dashboard passes the row's own `<text>` node and `getComputedTextLength()` — what gets measured is the real font size and weight, not a character-count estimate.
- Disambiguation is computed over every musician in the chart's cross-filtered data, not just the visible top 20, so a bare `Aaron` is unambiguous across the whole filtered range.
- The composers chart gets the ellipsis fallback but **not** the short form — `disambiguateLabels` is first-name/last-initial logic and would render `Vaughan Williams` as `Vaughan`.
- A measurement of 0 (the view is hidden, so nothing is laid out) reads as "fits" and leaves the full name; `notifyShown()` already re-renders at the real width.
- The literal `8` for the name/bar gap is now the named `NAME_GAP`, since the fitting math needs the same number.

## Testing

- `npm test` — 326 pass, including 6 new `fitText` cases (fits / falls back to short form / clips / skips empty candidates / treats an unmeasurable element as fitting / degrades to a bare ellipsis).
- `npm run lint` and `npm run typecheck` clean.
- Drove the built dashboard in Chromium against a synthetic fixture (placeholder names, since the real sheet and alias table are gitignored personal files). At 1280px only the genuinely long names shorten (`Francine Delacroix-Bell` → `Francine`) and the rest are untouched; at 390px most fall back to first names and the two first-name collisions resolve as `Chris O.` / `Chris B.`. No label's left edge goes negative at either width.

`CLAUDE.md`'s `DashboardComponent` entry documents the fitting ladder.

---
_Generated by [Claude Code](https://claude.ai/code/session_012DvTXSojoRFq2Uv7VWvw75)_


src/dashboardComponent.js
```diff
@@ -414,7 +423,29 @@ export class DashboardComponent {
...
+                        return node.getComputedTextLength();
+                    },
+                );
+                node.textContent = label;
+                // Native hover tooltip carrying the name we couldn't show.
+                // Set after textContent, which would otherwise drop it.
+                const titleSel = d3.select(node).selectAll('title')
+                    .data(label === d.name ? [] : [d.name]);
+                titleSel.exit().remove();
+                titleSel.enter().append('title').merge(titleSel).text(t => t);
```
jsundram: `<title>` is a hover affordance, and hover is exactly what's missing on the viewport where this fires most.

The mobile margin is 96px at a 14px font — roughly a dozen characters. Most full names in this log won't fit, so on a phone the musicians chart will shorten nearly every row, and `disambiguateLabels` shortens an unambiguous name to the **first name alone**. So the common mobile outcome is a Top-20 list of bare first names whose surnames have no path to the screen at all: no hover, and tapping the label calls `this.toggle(dimensionKey, d.name)` on line 425 instead of revealing anything.

That's arguably a worse regression than the bug being fixed. A name running off the SVG's left edge was at least partly readable; "Aaron" is complete, plausible, and wrong-by-omission — a reader can't tell it's been shortened.

I don't think `tooltip.attach` is a drop-in, and that's worth saying rather than hand-waving: its tap path would collide with the click-to-filter already bound to the same `<text>`. Options that don't collide, roughly in order of effort:

- Keep the full name in the label but let it use more room on mobile — bump `rankedMargin.left` in `sizing()` for the mobile branch and take it out of `right` (90px of gutter for a `12.3% (45)` label is generous).
- Derive the margin from the widest name actually in the top 20 (measure once, clamp to a fraction of `width`) so shortening is the exception rather than the mobile default.
- Keep shortening, but make it visible — a trailing `…` on *every* shortened form, not just the clipped one, so "Aaron…" reads as truncated where "Aaron" doesn't.

The `<title>` is worth keeping regardless for desktop and for screen readers; it just can't be the only answer. Worth deciding deliberately, since the PR description presents the short form as the good case and the ellipsis as the fallback, whereas on mobile the short form *is* the lossy one.

---

jsundram: The data-join here is doing no work. Line 441 assigns `node.textContent`, which replaces every child of the `<text>` — so by the time `selectAll('title')` runs on 444 the selection is always empty. `titleSel.exit()` is always empty too, and `merge(titleSel)` always merges nothing. The whole block collapses to:

```js
if (label !== d.name) d3.select(node).append('title').text(d.name);
```

which also says the intent (`the label is shortened, so carry the full name`) more directly than an enter/exit dance whose exit branch can't run.

The comment on 442–443 is the load-bearing part and should survive the simplification — "set after `textContent`, which would otherwise drop it" is precisely why this is safe.

---

src/dataProcessor.js
```diff
@@ -649,6 +649,36 @@ export function disambiguateLabels(nodes) {
...
+    }
+    const shortest = options[options.length - 1];
+    for (let n = shortest.length - 1; n > 0; n--) {
+        const clipped = shortest.slice(0, n) + '…';
+        if (measure(clipped) <= maxWidth) return clipped;
+    }
+    return '…';
+}
+
 /**
```
jsundram, line 709: This is a linear backward scan, and in the dashboard each `measure` call writes `textContent` and then reads `getComputedTextLength()` — a write/read pair that forces a synchronous layout every iteration. For a 16-character name that's up to 15 forced reflows for one row, and the loop only runs on rows where nothing fit, i.e. exactly the rows this feature exists for.

What makes it worth fixing rather than shrugging at is line 84 of `dashboardComponent.js`: `window.addEventListener('resize', () => this.render())`, undebounced. A resize drag or an iOS URL-bar show/hide fires a stream of events, each re-rendering both charts, each row now doing at least one forced layout and the clipped ones doing a dozen-plus. Before this PR `.text()` did no layout reads at all, so this is a new cost on a hot path.

The prefix length is monotonic in width, so a binary search gets the same answer in ~4–5 measurements:

```js
let lo = 1, hi = shortest.length - 1, best = '…';
while (lo <= hi) {
    const mid = (lo + hi) >> 1;
    const clipped = shortest.slice(0, mid) + '…';
    if (measure(clipped) <= maxWidth) { best = clipped; lo = mid + 1; }
    else hi = mid - 1;
}
return best;
```

Same contract, same tests pass, and it keeps the `'…'` degradation for `maxWidth: 0` that `fitText(['Aaron'], 0, measureChars)` pins. (Debouncing the resize handler would be the other half of this, but that's pre-existing and a separate change.)

---

src/dataProcessor.js
```diff
@@ -649,6 +649,36 @@ export function disambiguateLabels(nodes) {
...
+// A measure() that returns 0 (the element isn't rendered — a hidden view)
+// makes the first candidate "fit", which is the right degradation: the full
+// name, exactly as before, and the caller re-renders when the view is shown.
+/**
+ * @param {string[]} candidates preferred form first, shortest form last
+ * @param {number} maxWidth
+ * @param {(s: string) => number} measure
+ * @returns {string}
+ */
+export function fitText(candidates, maxWidth, measure) {
```
jsundram, line 680: The `@param` is narrower than both the implementation and the callers. `options = candidates.filter(c => c)` exists specifically to tolerate holes, `dashboardComponent.js:434` passes `[d.name, shortLabels ? shortLabels.get(d.name) : null]`, and the test asserts `fitText(['Beethoven', null], …)`. This file is `// @ts-check`, so the annotation is the contract — it just isn't being enforced against the call site because `dashboardComponent.js` isn't checked.

```js
 * @param {(string|null|undefined)[]} candidates preferred form first, shortest form last
```

`shortLabels.get()` returns `string | undefined`, so `undefined` belongs in there as well as `null`.

Two smaller things while you're in here:

The "shortest form last" invariant isn't guaranteed by the caller. `disambiguateLabels` returns the *full* name when even `First L.` collides, so `candidates` can be `[name, name]` — harmless (one redundant measurement, correct result), but it means the clip branch may be clipping something that isn't shorter than the first candidate. Worth a word in the doc that duplicates and non-strictly-decreasing candidates are tolerated, since a reader could reasonably take the invariant as load-bearing.

And on placement: `fitText` takes an injected DOM measurement callback, which makes it a view-layer helper living in the module CLAUDE.md describes as importing nothing and holding the data layer's pure functions. It *is* pure, and `disambiguateLabels` sets a precedent for label logic living here, so this is defensible — I'd just note it in the CLAUDE.md paragraph you're already updating, so the next person doesn't have to re-derive whether it belongs.

---

src/dashboardComponent.js
```diff
@@ -347,6 +350,12 @@ export class DashboardComponent {
             .attr('transform', `translate(${margin.left}, ${margin.top})`);
 
         const sel = this.state.selections[dimensionKey];
+        // Person names get a shorter fallback form; composer names don't
+        // (disambiguateLabels is first-name/last-initial logic, which would
+        // turn "Vaughan Williams" into "Vaughan"). Built from every musician
+        // in the chart's data, not just the visible top N, so a name that
+        // shortens to "Aaron" is unambiguous across the whole filtered range.
+        const shortLabels = dimensionKey === 'musician' ? disambiguateLabels(allData) : null;
```
jsundram, line 362: The measure-returns-0 degradation is the whole feature's failure mode, and it's the *default* state on the dashboard rather than an edge case: `index.html:148` is `<div id="dashboard" style="display: none">`, and `init()` calls `render()` while it's still hidden. So on first render every `getComputedTextLength()` is 0, every name "fits", and the chart is exactly the pre-PR overflowing one. Correctness rests entirely on `notifyShown()` firing a second render.

Two things I'd want pinned rather than reasoned about:

1. Is there a paint between the view becoming visible and `notifyShown()`'s re-render? If the `display` flip and the re-render aren't in the same task, the first frame of the dashboard shows names running off the edge and then snaps. Worth an actual look on device — it's the kind of thing that reads as a glitch rather than a bug and so never gets reported.

2. Nothing tests it. `test/dataProcessor.test.mjs` covers `fitText`'s contract thoroughly, including `() => 0`, but the wiring — that `notifyShown()` re-renders with real measurements — is untested in either suite. `e2e/smoke.spec.js` already navigates to `#dashboard` and asserts on the tiles; asserting there that no `text.ranked-name` extends past the SVG's left edge (or, more simply, that at least one row carries a `<title>` at a narrow viewport) would catch a regression that otherwise silently reverts to the current behavior. The failure is invisible precisely because the fallback *is* the old rendering.

Separately, on the choice of `allData` over `data` for `disambiguateLabels`: agreed it's the right input, but note the consequence isn't only stability — it also means changing the date filter can flip a visible row between `Aaron` and `Aaron J.` without the top 20 changing at all, because a second Aaron entered the range off-screen. That's the correct behavior (the label stays honest) and I wouldn't change it, but the comment currently sells it as stability, which is the opposite of what a user will observe.

---

jsundram (commented):
The underlying bug is real and the diagnosis is right: the name is right-anchored at `-8` inside a group translated by `margin.left`, and the outermost `<svg>` clips by default, so anything wider than 120px (96 on mobile) was hard-cut with no indication. Measuring on the row's own `<text>` after the font-size and font-weight attrs are set — rather than on a scratch node — is the detail that makes this trustworthy, and reusing `disambiguateLabels` instead of inventing a second shortening rule is the right instinct. `fitText` is genuinely pure and the injected `measure` makes it testable; the `measureChars` stand-in is a nice touch.

`npm test` (326 pass), `npm run lint` and `npm run typecheck` are clean on `d8aba47` here.

Five comments inline. Two I'd resolve before merge:

**The mobile case is the one the design doesn't serve.** At 96px and a 14px font the musicians chart will shorten nearly every row, and an unambiguous name shortens to the bare first name — so a phone shows a Top-20 of first names whose surnames are unreachable, since `<title>` needs hover and tapping the label toggles the filter. A truncated-and-marked name is honest; `Aaron` reads as complete. I'd rather see the mobile margin widened (or derived from the widest visible name) than have shortening be the mobile default, and I've sketched the options inline.

**Nothing pins the wiring.** `#dashboard` starts `display: none`, so the first render always measures 0 and always produces the old overflowing chart; the fix only exists because `notifyShown()` re-renders. That's the documented degradation, but it also means a regression in that path restores the exact bug this PR fixes, silently, with every unit test still green. An assertion in `e2e/smoke.spec.js` would close it.

The other three are smaller: the `<title>` data-join has a dead exit branch (`textContent =` already cleared the children), the ellipsis search is a linear scan of forced layouts on an undebounced resize handler where a binary search is ~4 measurements, and `fitText`'s `@param` says `string[]` while both the implementation and the callers rely on nullable entries.

---

src/dashboardComponent.js
jsundram: Fixed in e31f2d5 — collapsed to the plain append.

You're right that the exit branch can't run, and the reason is worth keeping in the comment rather than just deleting the dance, so I folded it in: the same `textContent` assignment that makes ordering load-bearing is also what guarantees there's never an existing `<title>` to update or remove.

```js
node.textContent = label;
// Native hover tooltip carrying the name we couldn't show.
// Appended AFTER textContent, which replaces every child of
// the <text> and would otherwise drop it — which is also why
// this needs no data-join: there is never an existing <title>
// here to update or remove.
if (label !== d.name) d3.select(node).append('title').text(d.name);
```

---

src/dataProcessor.js
jsundram: Fixed in e31f2d5 — took your binary search as written.

The undebounced-resize argument is what makes it worth doing, agreed: before this PR the label path did zero layout reads, so every forced reflow here is new cost on a handler that fires in streams. A 16-character name goes from up to 15 measurements to ~4.

The `maxWidth: 0` degradation is preserved — `best` initializes to `'…'` and no prefix ever fits, so `fitText(['Aaron'], 0, measureChars)` still returns a bare ellipsis. All six `fitText` tests pass unchanged, which is the useful signal that the contract didn't move.

I left the resize debounce alone as you suggested — pre-existing, and a separate change.

---

src/dataProcessor.js
jsundram: All three done in e31f2d5.

**Type.** Widened to `(string|null|undefined)[]`, `undefined` included for exactly the `shortLabels.get()` reason. That surfaced a real consequence: `candidates.filter(c => c)` doesn't narrow for tsc, so `shortest` became possibly-nullish and `npm run typecheck` failed. Built the array explicitly instead, which narrows and says what it's doing:

```js
/** @type {string[]} */
const options = [];
candidates.forEach(c => { if (c) options.push(c); });
```

**The invariant.** Documented that it's a preference order, not a length guarantee — `[name, name]` is a normal input, one redundant measurement, correct result.

**Placement.** Noted in the CLAUDE.md paragraph: it lives there because it's pure (the DOM measurement is an injected callback, not an import) and because `disambiguateLabels` set the precedent, flagged as the module's one view-layer helper so the next person doesn't re-derive it.

---

src/dashboardComponent.js
jsundram: Both addressed in e31f2d5, and the `allData` note too.

**1. No paint between the flip and the re-render.** This is settled by control flow rather than observation — `navigationComponent.js:155` `applyView()` does the `display` flip and then calls `onViewChange` in the same synchronous function body:

```js
d3.select(VIEW_TO_SELECTOR[view]).style("display", "block");
if (this.onViewChange) this.onViewChange(view);
```

and `app.js:138` `handleViewChange` calls `notifyShown()` → `render()` synchronously. The browser can't paint mid-task, so the 0-measurement render is never on screen — there's no frame to glitch. Worth saying plainly: I verified that from the code path and in headless Chromium, **not** on a physical device, so if you want the iOS check you flagged, that's still yours. I don't think there's a device-specific way for it to be false, though: it would take the flip and the call landing in different tasks.

**2. Pinned.** New test in `e2e/smoke.spec.js` at a 390px viewport asserting no `text.ranked-name` crosses the SVG's left edge, plus that the over-wide name carries its full name in a `<title>`. One fixture player gains a surname to be that name; KPI counts are untouched.

I checked it isn't vacuous — reverting the `.each` block to `.text(d => d.name)` and rebuilding fails it at `-17.6875px`, i.e. it catches exactly the silent revert you described.

**On `allData`.** You're right and I had the comment backwards — it was selling stability, which is the opposite of what's observable. Rewritten to say it buys honesty, and to name the consequence: widening the date range can flip a visible row from `Aaron` to `Aaron J.` with the top 20 unchanged, because a second Aaron entered the range off-screen. Correct in both cases.

Leaving this thread open for the device check.

---

src/dashboardComponent.js
jsundram: Agreed, and I'm leaving this open rather than picking for you — but I measured the options first, because the numbers rule one of them out.

**Confirming the diagnosis.** At a 390px viewport with a cast of full names, **9 of 9 rows shortened**. Shortening isn't the mobile fallback, it's the mobile rendering. And you're right that `<title>` can't answer it: tapping the label is bound to `toggle()`, so the surname has no path to the screen.

**Option 1 (take it out of `right`) doesn't get there on its own.** The widest count label measures 63.6px and ends 20.4px inside the SVG's right edge, so `rankedMargin.right: 90` has only ~20px of genuine slack. Full names at the 14px mobile font run ~120–135px and need ~136px of left margin. So ~20px comes from the right gutter and the rest has to come out of the bar: the plot area goes 204px → ~184px, about 10%. That's affordable, but it's a bar-width cost, not free gutter — worth stating plainly rather than discovering later.

**Option 2 (derive from the widest visible name) is what I'd build**, with one wrinkle you should know about: the two ranked charts currently share a left edge because they share a constant, and `renderStats()` reads `s.rankedMargin.left` inline to align the KPI tiles with the plot area. A per-chart derived margin breaks both alignments. So it wants to be one margin computed across *both* charts' data and fed to the tiles too — contained, but more than a knob change, and it makes the tile row's indent data-dependent.

**Option 3 is worth doing regardless, narrowed.** A trailing `…` on the bare-first-name form only — `Aaron…` — since `Aaron J.` already reads as abbreviated on its own and `Aaron J.…` is just noise. That closes the "complete, plausible, and wrong-by-omission" hole for whatever still shortens past the clamp, and it's a few lines.

My recommendation: **2 + narrowed 3**, clamping the derived margin to ~40% of chart width so a pathological name can't eat the bar, with shortening kept as the honest fallback past the clamp. Happy to build 1 + narrowed 3 instead if you'd rather keep the margin a constant — it's much the smaller diff.

Your call; I'll implement whichever you pick. Flagging that either way this is a mobile-layout change I can only verify in headless Chromium at a 390px viewport, not on a real phone.

---

jsundram: Decision: tracking this separately rather than widening #30 — filed as #31, with the measurements, the three options and the recommendation carried over so nothing has to be re-derived.

To be explicit about what that means for this PR: the overflow bug is fixed and no label crosses the SVG edge at any width, but on mobile a bare shortened first name still reads as complete. That's a known, tracked gap, not an oversight.

Resolving this thread on that basis; #31 is where the fix gets decided.

---

Issue #32, opened 2026-09-01
attribution: cover initialled names ("Zelda Q"), which no summary counts today


jsundram:
`attribution` takes subjects only where `nameShape(name) === 'bare'`, so an **initialled** name — `Zelda Q`, a surname abbreviated to a letter — is never a subject. The row could decide it exactly the way it decides a bare `Zelda`, and nothing else decides it either.

### Why it matters more than "one uncovered shape"

The evidence already exists and is already printed. On a fixture where `Zelda Quinton` is written out 8× (always with Beryl Adams + Carol Diaz) and `Zelda Quiller` 6× (always with Dexter Stone + Nora Vance), a row reading `Zelda Q | Beryl Adams | Carol Diaz` gets this from `audit_aliases`:

```
    ≠ ambiguous  'Zelda Q' [upper]  (overlap=100%, 1×) — could be 'Zelda Quiller' or
                 'Zelda Quinton'; no alias can pick, fix the rows (see AMBIGUITY)
  ✓ [upper] 'Zelda Q'  (1×)  →  'Zelda Quinton'  (8×, overlap 100%)
    [upper] 'Zelda Q'  (1×)  →  'Zelda Quiller'  (6×, overlap 0%)
```

Three things then go wrong together:

1. It says **"see AMBIGUITY"**, and the AMBIGUITY section is single-token only — the row it points at is not there. The cross-reference dangles.
2. `-- bare names in the sheet with 2+ candidates (0) --`, so the `audit_all.sh` SUMMARY line marked **NEEDS MEMORY** reads **0**. A routine monthly run reports nothing to answer.
3. `attribute()` returns **0 findings** for it, and does not count it in `settled` or `unverified` either — so the reassurance line ("N agree with the table, M have an alias standing") does not mention it. It is silently absent from every total.

These are the findings that decay. "Which Zelda was this" is answerable today and unanswerable in a year.

### Where the change goes

`scripts/attribution.mjs:128` — `if (nameShape(name) !== 'bare') continue;` is the one-word site (that is why `nameShape` exists). But candidate selection has to narrow too: `candidatesFor(byFirst, baseToken(name), cls)` keys on the first token alone, which for `Zelda Q` would also return `Zelda Xu` — a name `Zelda Q` does not abbreviate. The narrowing already exists at `scripts/audit_aliases.mjs:204`:

```js
.filter(c => abbreviates(variant.name, c))
```

Reuse it rather than writing a third rule.

### Cases

[...]

PR #33, opened 2026-09-05
Log a Piece: an in-app entry form that writes through your own Google Form

jsundram:
Adds a `#log` view: an entry form that writes a new row by POSTing to the user's own Google Form's `formResponse` endpoint. Forms stays the writer, so the sheet keeps its `Timestamp` column and every reader downstream — `fillForward`, the audits, the CSV export — sees rows it already understands. Only the UI is ours.

**Not deployed or field-tested yet.** See *Before merging* at the bottom.

## Why

Pinning the Google Form to an iOS home screen opens a new Safari tab on every launch: a Form ships no web app manifest and no `apple-mobile-web-app-capable` tag, so the pin is a bookmark rather than an app. This site ships both, so `log.quartetroulette.com/#log` launches standalone.

That was the prompt. The reason it's worth more than a tab fix is that the app has the whole log in memory at entry time and Google Forms does not.

## What the form knows that the Form can't

- **Composers you play are one tap**, ranked from your own data, with the full catalogue behind *More…* and free text behind that. The Google Form lists its seven as radios, so a 22-item picker would have been *slower* than the thing it replaces.
- **Names autocomplete from the log's canonical spellings** — the bare-first-name ambiguity `scripts/attribution.mjs` chases, prevented at the source instead of audited after.
- **Empty seats show who they'll repeat** as placeholder text: `howto.md` §6's ditto rule made visible rather than trusted.
- **Each seat has a part beside it.** Two people swapping is a dropdown, not retyping both names into different columns; a quintet's second viola or cello can be said outright. Written as the same `(code)` annotation the sheet has always used.
- **`Others?` gets the same control**, plus a freeform box for entries carrying prose (`Laura (v2, shadowing on I)`), and **extras carry through the sitting** — the sheet can't ditto that column, and `howto.md` §6 calls forgetting it the single most common way a person goes missing from the log. The **x** is how you say someone left.
- **People already in the sitting are one-tap chips**, carrying the instrument they were last logged on: the second sextet of an afternoon has the first one's players.
- **Everything but the send works offline**, submissions queue in order, and nothing on screen is lost to an eviction.

## Load-bearing decisions

**Which form is per-user config, with no default in the bundle.** A correctness constraint, not a preference: the site is deployed publicly and anyone can point it at their own sheet, so a hardcoded form id would land every visitor's entries in one person's spreadsheet — and they'd see "Logged", because the `no-cors` response is opaque. Entry ids come from a Google Forms **pre-filled link**, the only way to learn them in a browser. Ids map to columns positionally, correct by construction because Forms builds the sheet's columns from its questions in order; `LABELS[FIELDS[i]] === CSV_HEADERS[i+1]` is the invariant and a test pins it.

**A `?form=` link is a proposal, never adopted.** A URL is not a choice, and this config alone decides where every row is sent.

**Composer rides Forms' `__other_option__` escape; Which Part does not.** The escape is for unbounded value spaces only, and the asymmetry is the point. Composer is unbounded — the catalogue knows twenty and `Other…` accepts anything typed — so no option list could cover it and the app cannot read the user's list anyway. Which Part is a fixed four that a form feeding this sheet must already offer. That matters because the escape *requires* "Other" to be switched on, and a multiple-choice question without it rejects the whole response, silently: routing a four-seat question through it would have made every submission depend on a setting nobody would think to enable there. So the documented precondition is about Composer alone (`howto.md` §1 and the setup panel), and Which Part just needs V1/V2/VA1/VA2 among its options.

This replaced a hardcoded list of the reference form's seven composers, which sent those seven plainly and everything else through the escape — so a short-answer Composer question wrote seven correctly and put the literal `__other_option__` in the cell for the other thirteen, starting on whichever piece first used an unlisted one and invisible behind the opaque response. A form that doesn't meet the condition now fails on the first piece instead.

**`flush` is sequential, stops at the first failure, and is guarded against re-entrancy.** `fillForward` reads each row against the one above it, so an out-of-order flush would point a blank seat at the wrong previous row; two overlapping flushes would POST the same entry twice. Joining a flush is not returning it, either — that promise can settle a microtask before `inFlight` clears, and inheriting its answer means "Logged" for an entry still queued.

**Every localStorage write is checked.** The queue *is* localStorage, so an unpersisted entry is one `flush` will never see; `enqueue` says so and `submit` posts it directly rather than reporting a success nothing holds.

**Editing the sheet afterwards still wins.** The form only ever writes. The local copy of a submission expires by *identity*, never by the clock — Forms timestamps a row when it receives it and the device saves a moment later, so "is the sheet newer than my copy" is never true.

## Testing

526 unit tests (`node:test`), 37 Playwright tests, lint and typecheck clean.

`e2e/smoke.spec.js` is no longer only a boot smoke: most of it drives the form and asserts on the **submitted request body**, because that's where this feature's failures are visible. Three of the five defects the first code review found looked completely correct on screen — carried extras that displayed and submitted empty, a removed player who came back, a form whose part selection was cleared from another view.

## Review passes

Security, UX, code review and documentation, in that order, then a second review round on the finished branch. Each found something real:

- **Security** — `?form=` re-pointed a configured device in one click. My first fix guarded only devices that already had a form, which is nobody the day this ships; the review caught that too.
- **UX** — name fields were letting iOS autocorrect surnames, which is the exact failure the feature exists to prevent.
- **Code review** — four defects, all mine, including three copies of a stray paste from a careless global replace.
- **Documentation** — writing the privacy section accurately exposed that **Log Out** cleared only the data URL and CSV cache, leaving the queue, the sitting and the draft (all full of player names) plus the form config, which would have sent the next person's entries to the previous person's sheet.
- **Second round (15 threads, all resolved)** — a submit silently lost when localStorage refuses a write while the UI says "Logged"; a `flush` that resends forever when it can't clear an entry; the tail-window join above; a discarded queue entry still steering carry-forward for 12 hours; `Others?` seeded from raw text while the chips came from canonicalized names, so an aliased person was written in twice; carry-forward reading the app's rows rather than the sheet's; and eight smaller ones. Two of my first attempts at tests for these passed against the unfixed code — both were rewritten until they failed without the fix.
- **After that** — the fix for the `CHOICES` thread first sent *both* choice questions through the escape, which quietly made every submission depend on "Other" being enabled for Which Part. Corrected in `9eeec49`; a plain value there works whether that question is multiple-choice or short answer, with or without Other.

## Before merging

- **Confirm your form's Composer question is multiple-choice with "Other" enabled.** Every composer now rides the escape, so this is the one setting the whole submit path depends on. Related and unverified from here: that Forms accepts the escape for a value that *is* one of the question's options, storing it as plain text in the column. Log one piece with a listed composer and one with an unlisted one, then look at the sheet — both should read as ordinary names.
- **No real-device testing.** The e2e can't cover an actual iOS PWA, backgrounded mid-entry, on a flaky connection, writing to a real sheet. Worth logging a few throwaway pieces against a scratch form first.
- **Setup is one paste per device**: your form's *pre-filled link*. **Copy setup link** in the menu now carries both the sheet URL and the form config, so a second device is one link.
- **Cellists can't log from the app**, and `howto.md` §1 and the setup panel now say so. Pre-existing and unchanged — the Google Form has the same four options — and tracked as #34 rather than widened into this PR.
- Analytics for this view (which forms, how many submits) is #35.
- ~~pandoc isn't available in this sandbox~~ — resolved: pandoc now runs here, `./build.sh --prod` completes, and `about.html` / `howto.html` are confirmed to render.
- Unrelated: `md/about.md` links to `github.com/jsundram/musiclog` rather than `quartet-log`, and its Privacy section says traffic goes to Google and nowhere else, which the existing GoatCounter tag already contradicts. Both pre-existing; the second is noted in #35.

https://claude.ai/code/session_016FH5WZachqUVbvHd33YvBs

---
_Generated by [Claude Code](https://claude.ai/code/session_016FH5WZachqUVbvHd33YvBs)_


src/logStore.js
```diff
@@ -0,0 +1,220 @@
...
+/** @returns {Queued[]} */
+export function pending() {
+    const list = read(PENDING_KEY, []);
+    return Array.isArray(list) ? list : [];
+}
+
+/** @param {Entry} entry @returns {Queued[]} */
+export function enqueue(entry) {
+    const list = [...pending(), { id: `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`, at: Date.now(), entry }];
+    write(PENDING_KEY, list);
```
jsundram: **A submit is silently lost when `localStorage` cannot be written, and the user is told "Logged".**

`write()` returns a boolean, but `enqueue` ignores it and returns the in-memory `list` regardless. `submit()` never posts directly — it always goes `enqueue` → `flush` — and `flush` re-reads `pending()` from storage, which is still empty. Verified against the branch with a `setItem` that throws:

```
enqueue() return length: 1   persisted pending(): 0
flush result: {"sent":0,"remaining":0}  posts actually made: 0
=> status shown: "Logged Haydn 76#3."  (kind: ok)
```

The piece is never POSTed, never queued, and the green confirmation is indistinguishable from a real one. This is not hypothetical for this app: it caches the whole parsed CSV in localStorage and `statusBar` already has a storage-full state, so `QuotaExceededError` is a condition it expects. Private-mode Safari throws on every write.

`test/logStore.test.mjs` has `a blocked localStorage does not break the form`, which pins that `enqueue` does not *throw* — it does not check that the caller learns the entry was dropped.

Suggest having `enqueue` return the write result (or null) and `submit()` refuse to report success when the entry did not persist.

src/logStore.js
```diff
@@ -0,0 +1,220 @@
...
+/** @type {Promise<{ sent: number, remaining: number }>|null} */
+let inFlight = null;
+
+export async function flush(post) {
+    // One flush at a time. `online`, becoming visible and submit all call
+    // this, and two overlapping runs would each read the same head entry and
+    // POST it — a duplicate row in the sheet, which nothing downstream can
+    // tell from a real one. Joining the run in progress also means a submit
+    // during a flush is picked up by it, since the loop re-reads the queue.
+    if (inFlight) return inFlight;
```
jsundram: **Joining an in-flight flush can return an already-settled run, so `submit()` reports "Logged" for an entry that was never posted.**

The comment above says a submit during a flush "is picked up by it, since the loop re-reads the queue". That holds while the loop is still running, but not in the tail window: the IIFE can have finished its loop and computed `remaining` while `inFlight` is still non-null (it is only cleared in the first callers `finally`, one microtask later). A second caller that enqueues and then calls `flush` in that window gets the *previous* runs result.

Verified on the branch — an `online` flush finishing, then a tap on Log it in the same tick:

```
posted: []
submit() saw: {"sent":0,"remaining":0}
status shown to user: Logged Haydn 76#2.  It reaches the app within a few minutes.
actually still queued, never sent: [ 76#2 ]
```

The entry is not lost (the next `online` / `notifyShown` drains it), but the user is told it reached the sheet when it did not — and the pending list is only re-rendered from the same stale result. Reachable via the `online` listener firing with a non-empty queue while the user is mid-entry.

A cheap fix is to have the joiner chain a follow-up run rather than return the settled promise, e.g. keep looping while `pending()` is non-empty after the awaited run resolves.

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+            .text(n => `${n} waiting to send, in this order. The sheet timestamps them on arrival.`);
+        box.selectAll('.log-pending-row')
+            .data(queued, d => d.id)
+            .join(enter => {
+                const row = enter.append('div').attr('class', 'log-pending-row');
+                row.append('span').attr('class', 'log-pending-what');
+                row.append('button')
+                    .attr('type', 'button').attr('class', 'log-pending-drop')
+                    .attr('aria-label', 'Discard this entry').text('x')
+                    .on('click', (_, d) => { store.drop(d.id); this.renderPending(); });
```
jsundram: **Discarding a queued entry with the x leaves its `setRecent` copy behind, so a row that will never exist goes on driving carry-forward for 12 hours.**

`submit()` calls `store.setRecent(resolved)` unconditionally — including when the flush failed and the entry is still queued. The x here only calls `store.drop(id)`, which touches `PENDING_KEY`. Nothing removes the matching entry from `RECENT_KEY`.

Concretely: log a piece in a basement (queues), then decide it was a mistake and press x. `carrySource()` still returns that entrys `resolved` copy via `store.recent()`, so every seat placeholder, `slotCell`s blank-vs-write decision and `defaultOthersCell()` are computed against a row the sheet will never hold — until `RECENT_MAX_AGE_MS` (12h) expires it. `isSameRow` can never retire it either, because the sheet will never contain a matching composer+title.

The consequence is not cosmetic: `slotCell` returns `` (blank ditto) for any seat that matches the phantom carried cell, so the next real submission writes blanks that `fillForward` will resolve against a different row entirely.

`logStore` needs a way to forget a recent entry (or `setRecent` should only run once the POST has actually left the device), and the x should call it.

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+     * howto section 6 calls forgetting to the single most common way a person
+     * goes missing from the log. So the form carries them instead and writes
+     * them out each time; the x on a row is how you say someone left.
+     *
+     * Scoped to the sitting, unlike the seats: a blank seat repeats however
+     * long the break, but re-adding the people from three days ago would be
+     * plainly wrong.
+     */
+    defaultOthersCell() {
+        return sessionRows(this.sessionSource()).at(-1)?.others ?? '';
```
jsundram: **`Others?` is seeded from the RAW `others` cell while the session chips come from the canonicalized `othersList`, so an aliased person is offered a second time and submitted twice.**

`defaultOthersCell()` reads `row.others` — the untouched sheet string, which `normalizePlayerNames` deliberately leaves alone ("the raw `others` string stays untouched for the CSV-download path"). `sessionPeople()` reads `row.othersList`, whose names have been through `canonicalize`. The two disagree exactly where `PLAYER_ALIASES` does its job.

Verified on the branch:

```
defaultOthersCell is the RAW cell: "Pete (vc)"
seeded row name:                   ["Pete"]
sessionPeople name (canonical):    "Peter Ouyang"
=> chip "+ Peter Ouyang" is not in `taken` {Pete}; tapping yields:
   "Pete (vc); Peter Ouyang (vc)"
```

The same person lands in the cell twice, under two spellings, and the row now over-counts the ensemble — which is precisely what `audit_ensembles` and `attribution` exist to chase.

Note this can never fail in CI or in the e2e: `src/aliases.js` is the empty stub there, so raw and canonical are identical. It only fires on a real device with a populated alias table — the exact machine-vs-CI divergence CLAUDE.md flags under "Alias privacy" ("a test that read it would pass in both places while testing two different things").

The seats do not have this problem, because `carriedForward` reads the canonical `player1..3`. Seeding the `Others?` rows from `row.othersList` (re-serialised) rather than `row.others` would make the two views agree.

src/formConfig.js
```diff
@@ -0,0 +1,211 @@
...
+// value outside a radio's option list has to arrive through Forms' "Other"
+// escape: the entry carries a sentinel and the real text rides on a companion
+// field. We cannot know another user's option lists without fetching their
+// form (cross-origin, so we can't), and Other is harmless on a question that
+// accepts free text anyway — so these are the values known to need it, and
+// anything unlisted is sent through the escape rather than guessed at.
+export const CHOICES = {
+    composer: ['Bartok', 'Beethoven', 'Boccherini', 'Haydn', 'Mendelssohn', 'Mozart', 'Shostakovich'],
+    part: ['V1', 'V2', 'VA1', 'VA2'],
+};
```
jsundram: **`CHOICES` is the reference forms option list, applied to every users form — on a short-answer question it writes the literal `__other_option__` into the sheet.**

The whole design premise of this module is that the form is per-user and unknowable ("We cannot know another users option lists without fetching their form"). `toFormBody` nonetheless routes any value outside *this* hardcoded list through Forms `__other_option__` escape for every configured user.

The comment argues the escape is "harmless on a question that accepts free text anyway", but that is not how Forms behaves. For a **short-answer / paragraph** question, `entry.N.other_option_response` is ignored and the value of `entry.N` is stored verbatim — so the Composer column gets the string `__other_option__`. For a radio question that does **not** have "Other" enabled, the whole response is rejected, and `mode: no-cors` means the app still says "Logged".

Since the setup flow accepts any pre-filled link with the right field count, a user whose Composer question is a text field (a perfectly natural way to build this form, and one `md/howto.md` §1 does not forbid) gets a corrupted or missing Composer on every piece that is not one of these seven names.

It also bites the reference form: `all_works.json` ships `Brahms`, `Dvorak` and `Schumann`, none of which are in `CHOICES.composer`, so three catalogued composers always take the escape path.

`CHOICES.part` has the same shape problem *and* doubles as the log forms visible part buttons (`buildPartButtons` reads it), coupling the UI to one users radio options.

This feels like the wrong depth for the fix: rather than a fixed list, either capture each questions type/options at setup time (the pre-filled link cannot give it, but the user could be asked), or send the plain value and let Forms decide — a rejected radio value is no worse than a corrupted text one, and is at least consistent.

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+     * loses by putting the phone down.
+     */
+    restoreDraft() {
+        const draft = store.readDraft();
+        if (!draft) return this.seedOthers();
+        this.entry = blankEntry(draft.entry);
+        this.slotPartOverrides = draft.slotPartOverrides ?? [null, null, null];
+        this.othersFree = draft.othersFree ?? '';
+        this.expandComposer = !!draft.expandComposer;
+        this.otherRows = (draft.otherRows ?? []).map(r => this.newOtherRow(r));
```
jsundram, line 649: **`restoreDraft` preserves the saved row ids but restarts `_otherId` from 0, so a later Add person can mint a duplicate d3 key and the new row silently never appears.**

`newOtherRow(row)` spreads `...row` last, so a saved `id` overrides the freshly generated one. The counter still advances once per restored row, which leaves `_otherId` at `otherRows.length` rather than at the maximum id in use. If any row was removed before the draft was saved, the ids are sparse and the counter is behind them.

Verified on the branch (add 5 extras, remove the first 3, get evicted, reload, add 2 more):

```
saved draft ids: [ 4, 5 ]
after restore:   rows [ 4, 5 ]   _otherId = 2
after adding two: [ 4:D, 5:E, 3:F, 4:G ]
DUPLICATE d3 key? true
```

`renderOtherRows` joins on `d => d.id`, so the second row with id 4 is dropped by the keyed join: the user taps `+ Add person`, nothing appears, and the focus call lands on the wrong input. Worse, the row *is* in `this.otherRows`, so `syncOthers` writes it into the `Others?` cell and it gets submitted — an invisible extra player on the row.

Either drop the saved id and re-key on restore, or seed the counter:

```suggestion
        this.otherRows = (draft.otherRows ?? []).map(r => this.newOtherRow(r));
        this._otherId = Math.max(this._otherId, ...this.otherRows.map(r => r.id ?? 0));
```

src/logEntry.js
```diff
@@ -0,0 +1,442 @@
...
+ * @param {string|null} a.chosen selected part key, or a raw code passed through
+ * @param {string|null} a.implied the part the seat implies
+ * @returns {string} the cell to submit
+ */
+export function slotCell({ typed, carried = '', chosen, implied }) {
+    const written = (typed ?? '').trim();
+    // "-" is "nobody in this seat" (howto section 5), not a person to annotate.
+    if (written === '-') return '-';
+
+    const name = written || stripParens(carried) || '';
```
jsundram, line 297: **`slotCell` guards a typed `-` but not a carried one, so changing a seats part over an empty seat writes `- (vc2)` into the sheet.**

Line 269 correctly refuses to annotate a `-` the user typed. But two lines down, `name` falls back to `stripParens(carried)`, and `stripParens("-")` is `"-"` — so the same value reaches the annotation branch through the carry path.

Verified on the branch:

```
slotCell({typed:"", carried:"-", chosen:"VC2", implied:"VC"})  ->  "- (vc2)"
slotCell({typed:"", carried:"-", chosen:"VC",  implied:"VC"})  ->  ""        (correct)
```

Reproduction: the previous row was a string trio, so Player 3 is `-`. The next piece is a quintet, so the user opens seat 3s dropdown and picks VC2 without typing a name — the "a part changed on a blank seat materialises the carried name" rule fires on `-`.

Downstream, `peopleKeysFor` skips only an exact `"-"`, so `- (vc2)` counts as a person named `-`; `instrumentFromSlot` reads it as a cellist; `stripParens` leaves `-` as the name. It becomes a phantom player in the unique-people stats and the part breakdown.

The guard wants to move below the carry fallback, or be repeated on `name`:

```suggestion
    const name = written || stripParens(carried) || ;
    if (!name || name === -) return written;
```

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+    // Everyone already in this sitting who is not already on the row --
+    // seats included, since a person moves between a seat and Others? as the
+    // ensemble changes. Tapping one brings the instrument they were last
+    // logged on, so the second sextet costs one tap per extra player.
+    renderSessionPeople() {
+        const carried = carriedForward(this.carrySource());
+        const taken = new Set([
+            ...SEATS.map(f => stripAnnotation(this.entry[f].trim() || carried[f])),
+            ...this.otherRows.map(r => r.name.trim()),
+            ...splitOthersCell(this.othersFree).rows.map(r => r.name),
```
jsundram: **`taken` drops every `Others?` entry carrying a comment, so a person already in the freeform box is still offered as a chip and gets written into the cell twice.**

`splitOthersCell(...).rows` filters out entries with a comment by design — they live in `freeform`. Using `.rows` here means the freeform half contributes nothing to `taken`, while `sessionPeople` (reading `row.othersList`, i.e. `parseOthers`, which keeps the person and discards only the comment) does offer them.

Verified on the branch:

```
seedOthers  -> rows: []   freeform: "Laura (v2, shadowing on I)"
`taken` contribution from the freeform box: []
sessionPeople sees: ["Laura"]
=> chip offered though she is already on the row: true
cell after tapping the chip: "Laura (v2); Laura (v2, shadowing on I)"
```

Laura is counted twice by `peopleKeysFor` and `audit_ensembles` sees a row one player too large — and the user has no reason to suspect it, because the chip explicitly claims she is not on the row yet.

`parseOthersRows(this.othersFree).map(r => r.name)` gives every name in the box regardless of comment, which is what this set wants:

```suggestion
            ...parseOthersRows(this.othersFree).map(r => r.name),
```

(needs `parseOthersRows` added to the `./logEntry.js` import list)

src/logStore.js
```diff
@@ -0,0 +1,220 @@
...
+
+// Is this fetched row the submission we saved? Composer and work title only:
+// both are required, so both are always written, while `part` is folded on the
+// way in (processRow turns VA1 into VA) and comparing it would report a
+// mismatch for every viola row. A false match only means preferring the sheet,
+// which is the safe direction; a false MISS is the one that hurts, because it
+// leaves the local copy shadowing the sheet.
+/** @param {any} row @param {Entry} entry */
+function isSameRow(row, entry) {
+    return row?.composer === entry.composer && row?.work?.title === entry.title;
```
jsundram, line 189: **`toFormBody` trims but `setRecent` does not, so a stray space in the title makes `isSameRow` a permanent false miss — the exact failure this doc comment says is the one that hurts.**

The submitted value is `(entry[field] ?? "").trim()` in `toFormBody`, and `processRow` trims again on the way back in. But the copy handed to `setRecent` comes from `resolveCarry(entry, carried)`, and `resolveCarry` only trims the four `CARRIED` fields — `title` and `composer` are copied verbatim by `blankEntry`.

So type `76#3 ` (trailing space, or paste one), and the sheet holds `76#3` while the remembered entry holds `76#3 `. `isSameRow` compares them with `===`, never matches, and `recent()` goes on returning the local copy until `RECENT_MAX_AGE_MS` expires it.

Per the comment right above: "a false MISS is the one that hurts, because it leaves the local copy shadowing the sheet" — and per the module doc, that is what defeats "correcting a name in the sheet afterwards is the established fix for a typo or a surname learned later". The e2e `a name fixed in the sheet afterwards wins over what was typed` passes only because its fixture types no stray whitespace.

The same applies to `composer` when it comes from the free-text `#logComposerOther` input.

Trimming in `resolveCarry` (or normalising in `isSameRow`) closes it; `test/formConfig.test.mjs` already pins the trim on the write side, so the two sides should agree.

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+     */
+    defaultOthersCell() {
+        return sessionRows(this.sessionSource()).at(-1)?.others ?? '';
+    }
+
+    // The row the sheet will read this one against. A submission this device
+    // made minutes ago beats the fetched data, which lags by however long the
+    // published CSV takes to catch up.
+    carrySource() {
+        const last = this.rows.at(-1) ?? null;
```
jsundram: **`carrySource()` reads the last row the *app* kept, but `fillForward` in the sheet reads the last row the *sheet* holds — and `processData` drops partial-movement rows.**

`this.rows` is post-`processData`, which filters out every title containing `:` (and every unparseable-timestamp row). Those rows are still in the spreadsheet, and they are what a blank cell on the next row will actually ditto.

So: log `Op. 59#1: I` (the form even warns you the app will hide it), then log the next piece with a seat left blank. The placeholder, and `slotCell`s decision to leave the cell blank, are computed against the row *before* the partial. If the partial changed a seat, the sheets `fillForward` resolves that blank to a different person than the form displayed.

Within one session `store.recent()` masks this, because the local copy shadows whatever the fetched rows say. It is exposed when the partial was logged from the real Google Form or from another device, or after the 12h `RECENT_MAX_AGE_MS` window — i.e. the carry is wrong precisely when the local copy is not there to paper over it.

`warnings()` already tells the user a `:` title vanishes from the app; it does not warn that the row still steers the next rows carry-forward. At minimum this is worth a comment naming the gap; better would be to keep the dropped rows available for the carry computation.

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+            // NUMBER of fields means the form doesn't match the ten columns
+            // this app requires, and re-pasting will never help.
+            d3.select('#logSetupError').text(SETUP_ERROR[read.reason] ?? '');
+            this.renderSetupMap();
+        });
+        d3.select('#logSetupSave').on('click', () => {
+            if (!this.pendingConfig) return;
+            setFormConfig(this.pendingConfig);
+            this.config = this.pendingConfig;
+            this.pendingConfig = null;
```
jsundram, line 165: **After Connect form, the Save button is left enabled and the mapping table is left rendered — so the setup panel reopened by Change shows the previous forms column mapping under an empty input, with a live-looking button that does nothing.**

This handler clears `pendingConfig` and the input value, but never re-disables `#logSetupSave` (which the `input` handler had enabled) and never calls `renderSetupMap()` to clear `#logSetupMap`. Both are only hidden by `renderMode()`, not reset.

Sequence: connect a form → tap **Change** (`#logChangeForm` clears the config and re-renders) → the setup panel comes back with the nine `Composer → entry.NNN` rows of the form you just disconnected still listed, and **Connect form** enabled. Tapping it hits `if (!this.pendingConfig) return;` and silently does nothing.

That stale mapping is the one thing this panel exists to let a user verify, so showing a disconnected forms mapping is worse than showing none.

```suggestion
            this.pendingConfig = null;
            d3.select(#logSetupLink).property(value, );
            d3.select(#logSetupSave).property(disabled, true);
            d3.select(#logSetupError).text();
            this.renderSetupMap();
```

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+
+    setOtherRows(rows) {
+        this.otherRows = rows.map(r => (r.id ? r : this.newOtherRow(r)));
+        this.syncOthers();
+        this.renderOtherRows();
+    }
+
+    // The cell is derived from the rows. Nothing writes back into them while
+    // the user is typing, so a background revalidate cannot move a caret.
+    syncOthers() {
```
jsundram, line 677: **Every keystroke in an `Others?` name field does two `localStorage` read+parse round trips, one write, and a full copy of the entire dataset array.**

`syncOthers` is the `input` handler for each Others row (line 624) and for the freeform box (line 356), so this runs per character:

- `touch()` → `saveDraft` → `JSON.stringify` + `localStorage.setItem`
- `renderSessionPeople()` → `carriedForward(this.carrySource())` → `store.recent()` → `recentList()` → `getItem` + `JSON.parse`
- `renderSessionPeople()` → `sessionPeople(this.sessionSource())` → `sessionSource()` spreads `[...this.rows, ...]` — a fresh copy of the whole log (thousands of rows) — plus a second `getItem` + `JSON.parse` via `recentAll()`, then `sessionRows` walks it and `splitOthersCell` re-parses every remembered submissions `others` cell.

`localStorage` is synchronous and main-thread; this is the phone-in-a-rehearsal-room path the whole feature is for.

`refresh()` compounds it: `renderPlaceholders`, `slotParts` and `renderSessionPeople` each call `carriedForward(this.carrySource())` independently — three separate storage reads for one value — and `renderFields` calls `frequentComposers(this.rows)` after `renderComposerChips` already did (two full scans).

Caching `carrySource()`/`sessionSource()` per render pass and invalidating them in `setData`/`submit` would remove nearly all of it. `sessionSource()` in particular only needs the tail of `this.rows` inside the session window, not a copy of the whole array.

src/logComponent.js
```diff
@@ -0,0 +1,754 @@
...
+    location: '#logLocation',
+    comments: '#logComments',
+};
+
+const CARRIED_INPUTS = ['player1', 'player2', 'player3', 'location'];
+const SEATS = ['player1', 'player2', 'player3'];
+
+// The name half of a cell that may carry an "(instrument)" annotation.
+const stripAnnotation = (/** @type {string} */ cell) =>
+    (cell ?? '').replace(/\s*\([^)]*\)\s*$/, '').trim();
```
jsundram: **`stripAnnotation` re-implements `dataProcessor.stripParens`, which this feature already imports.**

`logEntry.js` imports `stripParens` from `dataProcessor` and uses it inside `slotCell` for exactly this job — pulling the name half off a cell that may carry an `(instrument)` suffix. `logComponent` defines a second, subtly different copy and uses it in `sessionSource()` and in `renderSessionPeople`s `taken` set.

They are not equivalent: `stripParens` requires a non-empty parenthetical *and* a name before it (`/^(.+?)\s*\(([^)]+)\)\s*$/`, returning the input unchanged otherwise), while this regex strips a trailing `()` and can reduce a cell to the empty string. So the value that goes into `taken` here can differ from the value `slotCell` compares against `carried` at submit time — the two places that have to agree about whether a person is already on the row.

Given CLAUDE.mds standing point that a second copy of a shared parse is how the form and the app come to disagree (the reason `SLOT_TO_PART` and `splitOutsideParens` were exported rather than copied), this should import the existing helper:

```suggestion
import { stripParens } from ./dataProcessor.js;
```

and drop `stripAnnotation` in favour of it at lines 407-409 and 521.

src/formConfig.js
```diff
@@ -0,0 +1,211 @@
...
+    if (!text) return { config: null, reason: 'empty' };
+
+    let url;
+    try { url = new URL(text); } catch { return { config: null, reason: 'not-a-form-link' }; }
+    // The id is interpolated into the URL formAction builds, so it has to be
+    // an id and not arbitrary path text. Anchored at the segment end so an
+    // unexpected shape is REJECTED rather than silently truncated into a
+    // different form's address.
+    const formId = url.hostname.endsWith('google.com')
+        ? url.pathname.match(/\/forms\/d\/e\/([\w-]+)(?:\/|$)/)?.[1] : null;
```
jsundram, line 101: **`hostname.endsWith("google.com")` matches `evilgoogle.com` — the check wants a dot boundary.**

`"evilgoogle.com".endsWith("google.com")` is `true`, so a link like `https://evilgoogle.com/forms/d/e/ATTACKER_ID/viewform?entry.1=&…` parses as a valid pre-filled link.

The blast radius is limited today, and deliberately so: `formAction` hardcodes `docs.google.com`, and the id is constrained to `[\w-]+` and anchored at the segment end (as the comment says), so the only thing an attacker-shaped host gets them is a proposal for a `docs.google.com` form id — which `consumeFormParam` already refuses to adopt without a tap. So this is a hardening nit, not an exploit.

It is still worth fixing while the guard is being written, because the containment is incidental: any future change that took the *host* from the link rather than only the id would inherit an open redirect from a check that reads as if it validated one.

```suggestion
    const formId = (url.hostname === google.com || url.hostname.endsWith(.google.com))
        ? url.pathname.match(/\/forms\/d\/e\/([\w-]+)(?:\/|$)/)?.[1] : null;
```

CLAUDE.md
```diff
@@ -90,6 +90,7 @@ The SPA has three in-page views and one external page, all reachable from the ha
 - **`#main`** (Home) — composer tabs, filterable lists, sortable per-composer data tables, ALL tab with aggregate stats + flat table. Tab order/content is data-driven from `all_works.json`: single-composer keys plus multi-composer tabs (`5+`, `MISC`), whose catalog entries are `{ composer: titles[] }` arrays and whose work labels are composer-prefixed (`Mozart-K515`). Adding another multi-composer tab is a JSON-only change (`isMultiComposerTab` detects the shape).
 - **`#calendar`** — GitHub-contributions-style year grid; per-year stats column; "Last 365 days" header; per-day tooltips.
 - **`#dashboard`** — cross-filter charts: stacked part bar (V1/V2/VA) + horizontal top-composers bar chart. Clicking one filters the other.
+- **`#log`** — the entry form. Writes a new row to the sheet by POSTing to the user's own Google Form's `formResponse` endpoint, so Forms stays the writer and nothing downstream changes. Which form is per-user config (see `formConfig` below), so an unconfigured visitor gets a setup panel rather than someone else's form. Static markup in `index.html`; everything but the send works offline.
```
jsundram, line 93: **The sentence introducing this list still says "three in-page views" (line 88), but this PR makes it four.**

> The SPA has three in-page views and one external page, all reachable from the hamburger menu:

`#log` is the fourth, and `IN_PAGE_VIEWS` in `navigationComponent.js` now has four entries. The count wants updating alongside the bullet.

Same paragraph, one line up from the hunk — GitHub will not let me anchor a comment there directly.

While here: `md/about.md`s Privacy section closes with "**Log Out** clears the saved URL and the cached data from your browser", which is now the smaller half of what it does. The paragraph above it correctly enumerates the form config, the queue, the sitting and the draft — the closing sentence should say Log Out clears all of it, since that is the sentence someone reads before handing over their phone, and the PR body names this section as the one whose rewrite exposed the gap.

src/logStore.js
jsundram: Confirmed and fixed in c3b619c.

`enqueue` now returns `null` when the entry did not persist, and `submit()` posts it itself — after the flush, so anything that *did* persist still reaches the sheet first — reporting a failure and keeping the piece on screen when that fails too. There is no outbox holding it and no later attempt coming, so a transport failure has to be visible.

The same trap was one line down in `flush`: a failed removal leaves the entry in the queue the loop re-reads, so it would resend forever. That turned out to be more than theoretical — the first negative-control run pinned a core at 1.7 GB before I killed it, which is also why the test bounds itself and fails rather than hangs.

Three `logStore` tests and two e2e cases, all verified to fail without the fix. The e2e reproduces your trace exactly: `log-status ok` for a piece never posted.

---

src/logStore.js
jsundram: Right, and my comment was wrong: re-reading the queue covers the *loop*, not the tail after it. Fixed in df00e0c — a joiner waits for the run and then looks at the queue again, starting its own run if anything is still there.

Worth recording how narrow it is. My first test guessed at the tick count and passed against the unfixed code, which made it worthless. The window is exactly one microtask wide and sits at tick 1, so the test now sweeps the whole settle sequence and asserts the invariant at each: a resolved flush never claims an empty queue while one is still waiting.

---

src/logComponent.js
jsundram: Fixed in df00e0c. `logStore.forgetRecent(entry)` is the counterpart to `drop`, and the x calls both, then redraws the data-driven parts (not the field values — a repaint there would move the caret).

I took your first option rather than deferring `setRecent` until the POST leaves: the offline case is the one the sitting most needs remembering, since the published CSV lags by minutes and a basement session logs several pieces before anything reaches the sheet. Withholding the copy until the send succeeded would make carry-forward blind for exactly that stretch.

The e2e goes as far as the consequence you named: after the x the placeholder falls back to the sheet, and the next piece submits a written seat rather than a blank that `fillForward` would resolve against a row that never existed.

---

src/logComponent.js
jsundram: Fixed in df00e0c, though not quite as suggested: re-serialising `othersList` would have dropped the comment half, which `parseOthers` discards and this editor deliberately round-trips.

`canonicalOthersCell` (now in `logEntry.js`, since it is pure) keeps the raw cell for its comments and replaces only the names, positionally — both parsers split the same string with the same paren-aware boundaries and the same filter, so entry *i* is entry *i*. A length mismatch falls back to the raw cell rather than guessing which name maps to which.

Your note that this can never fail in CI is why the test states both views explicitly as a fixture (`others: 'Pete (vc)'` against `othersList: [{ name: 'Peter Ouyang' }]`) instead of relying on a table it will never have.

One thing your trace surfaced on the way: `sessionSource` was building each local submission's `othersList` from `splitOthersCell(...).rows`, which drops the commented entries entirely — so a person who was only ever in the freeform box was forgotten by the sitting. It now uses `parseOthersRows`, matching what `parseOthers` does to a fetched row.

---

src/logComponent.js
jsundram: Fixed in df00e0c with your second suggestion — seeding the counter, so restored rows keep the ids the draft was written against.

The symptom is not the one either of us predicted, and it took a negative control to find that out. d3 does render all four rows with a duplicate key, and the cell is correct at that point, so my first test passed against the unfixed code. What actually breaks is removal: rows are *removed* by id (`filter(r => r.id !== d.id)`), so with ids `[4, 5, 3, 4]` one x takes two people off the row — and not the one that was tapped:

```
DRAFT AFTER ADDS  ids [4, 5, 3, 4]  cell "Danilo; Elena; Fabio; Gisela"
click x on Danilo (id 4)
AFTER            rows 2             cell "Elena; Fabio"
```

Gisela is gone and nobody touched her. The e2e now pins that.

---

src/logEntry.js
jsundram: Fixed in df00e0c, with your suggested edit: the guard moves below the carry fallback so it catches `-` arriving by either route.

```
slotCell({typed:"", carried:"-", chosen:"VC2", implied:"VC"})  ->  ""   (was "- (vc2)")
```

Both branches pinned in `test/logEntry.test.mjs`.

---

src/logComponent.js
jsundram: Fixed in df00e0c with your suggested edit — `parseOthersRows(this.othersFree)`, with the import added.

My first e2e for this passed against the unfixed code, because the fixture's only `Others?` row is eight days old and the sitting was therefore empty: I was asserting that zero chips is zero chips. It now submits a piece with the person in `Others?` first, so she is genuinely in the sitting, then moves her to the freeform box with a comment and asserts no chip — which does fail without the fix.

---

src/logStore.js
jsundram: Fixed in df00e0c. `resolveCarry` now trims every field before the carried ones are resolved, so the remembered copy is what was submitted rather than what was typed.

Trimming there rather than normalising in `isSameRow` because the copy is also what carry-forward and the placeholders read: a title with a trailing space would otherwise be displayed and compared in two different shapes. `toFormBody` trims on the way out and `processRow` on the way back in, so this is the third side of the same agreement.

---

src/logComponent.js
jsundram: Took the better option rather than the comment, in df00e0c: `processData` now returns `{ rows, sheetRows }` — the rows every view shows, and every row the sheet holds — and `LogComponent` uses `sheetRows` for `carrySource()` and `sessionSource()` while suggestions stay on the visible rows.

It stayed small: two call sites in `app.js` and no other caller. The people on a partial-movement row were in the room, so the sitting should offer them back too, which the same change gets.

`warnings()` is left alone deliberately. It still says a `:` title vanishes from the app, which is true; it no longer needs to warn that the row steers the next one's carry, because it no longer does so invisibly.

---

src/logComponent.js
jsundram: Fixed in df00e0c with your suggested edit, and pinned by an e2e that connects a form, taps **Change**, and asserts the panel comes back empty: no value, no mapping rows, Connect disabled, no error text.

Agreed on the reasoning — a mapping table is only worth showing if it describes the form you are about to connect.

---

src/logComponent.js
jsundram: Fixed in df00e0c by caching, which is your first suggestion and not the second.

`carrySource()`, `sessionSource()` and `frequentComposers` are memoised per data change, with `invalidateSources()` as the single seam — called from `setData`, from `submit` after `setRecent`, and from the pending-list x. `carried()` memoises `carriedForward` on top, so the three renderers that each asked for it independently now share one read. A keystroke does no `localStorage` work at all beyond the draft write, which has to happen.

I did not narrow `sessionSource()` to the session tail. The tail cannot be taken by a simple slice — `sessionRows` chains by the gap between consecutive rows, so a long sitting reaches arbitrarily far back — and reproducing that rule in the component to bound the copy is exactly the kind of second implementation that keeps going wrong here. With the memo the copy happens once per fetch or submit instead of once per character, which is the part that was actually costing anything.

While in there: `redrawFromData()` now names the set of renders that are safe on a data change, so `setData` and the x share one list rather than two that can drift.

---

src/logComponent.js
jsundram: Fixed in df00e0c — `stripAnnotation` is gone and both sites use the imported `stripParens`.

The difference you point out is the one that mattered: `taken` and `slotCell` have to agree about whether a person is already on the row, and they were reducing the cell with two different regexes.

---

src/formConfig.js
jsundram: Fixed in df00e0c with your suggested edit, plus a test case pasting an `evilgoogle.com` link with a full set of entry ids.

Agreed on the framing: the containment is incidental, and a check that reads as if it validated the host should do it.

---

CLAUDE.md
jsundram: Both fixed in df00e0c: the count says four, and `md/about.md`'s Privacy section now closes by naming everything **Log Out** clears rather than the smaller half. I also fixed the same understatement in the menu description further up the page, which had the identical wording.

You are right that it is the sentence someone reads before handing over their phone — that paragraph is what exposed the gap in the first place, so it should not go back to describing only part of the fix.

---

src/formConfig.js
jsundram: Agreed on the diagnosis, and I have **not** changed the submit shape — it needs a decision I should not make for you.

What I did fix in df00e0c is the half that needs no decision: the visible Part buttons now read `logEntry.PART_CHOICES` (the seats a logger can say they were on) instead of `CHOICES.part`. Same four values today, which is exactly why the coupling was invisible; they are now pinned apart so a change to the submit shape cannot silently change what anyone can log.

The submit shape itself is a genuine fork, because the two forms are mutually exclusive and the app cannot tell which it faces:

- **Radio with Other enabled** — `entry.N` must be `__other_option__` or the response is rejected outright.
- **Short answer** — `entry.N` is stored verbatim, so the same request writes the literal `__other_option__` into the Composer column.

So your option (b) — send the plain value — is not strictly safer, it just moves the damage: it fixes the text-question user and breaks your own form for Brahms, Dvorak and Schumann, which would be rejected and, per the opaque response, still say "Logged". A rejected row and a corrupted one are equally silent, and I do not think one is obviously better than the other.

That points at (a): ask once at setup, since the pre-filled link cannot tell us. Concretely — a checkbox per radio-shaped question in the setup panel ("Composer is a multiple-choice question"), defaulting to *text*, stored beside the entry ids. Text is the safer default: it writes the wrong-but-visible value rather than dropping the row, and it is what a form built by someone following `howto.md` §1 most likely is.

Three things I would want your call on before building it:

1. Is your own Composer question a radio with Other enabled? That is the case the escape exists for, and it decides whether the default can be *text* without breaking you.
2. Is one checkbox per question acceptable in the setup panel, or does that make setup too heavy for the one-paste flow?
3. Does `Which Part` need the same treatment, or is it always going to be four fixed options?

Happy to implement whichever you prefer — it is contained to `formConfig` and the setup panel — but I would rather not guess about where other people's rows land.

---

jsundram: Resolved in abd7fba, and closer to your first framing than to either of the options I proposed.

`CHOICES` is deleted. Composer and Which Part now ride the escape **always**, not just for unlisted values — which needs no knowledge of a form we can't read, since Forms stores an Other response as plain text in the response-sheet column, so a value that *is* an option lands in the same cell. The only visible difference is the Form's own summary view, which nothing here reads.

That trades a partial failure for a total one, deliberately. Before: on a short-answer form, seven composers wrote correctly and thirteen corrupted, starting on whichever piece first used an unlisted name — delayed, intermittent, and invisible behind the opaque response. Now such a form fails on the first piece, which is the one someone actually checks.

That only works if the condition is stated, and nothing stated it: howto §1 said "fill it out with the questions you'd like to track" beside a screenshot of one form. It now names both requirements, says what each wrong shape looks like in the sheet (`__other_option__` in the cell, or a rejected row), and tells the reader to log one piece and look. The setup panel repeats it where a form is connected, since neither condition is checkable from here.

Also now said out loud, in both places: **a cellist can't log from the app.** `SLOT_TO_PART` has no `VC` key and `SLOT_CLASS` hardcodes slot 3 as the cello, so a cellist's row would put a violist in the cello slot — aliased in the wrong class and counted in the wrong column. `PART_CHOICES` therefore offers no VC button, which makes the failure "can't log" rather than "logs rows the reader misreads". Pre-existing and unchanged, since the Google Form has the same four options; it just wasn't written down. Happy to file it as an issue alongside #32 if you want it tracked.

One thing I have **not** verified empirically, and it is the load-bearing assumption: that Forms accepts the escape for a value that is itself an option. Worth one throwaway form before merging.

---

